PR #1, opened 2023-10-27
Improvements of instructions

petitfou:
- Add grub-update command in setup-host.sh to make the kernel argument effective
- Improve README
- Remove useless QEMU argument in run_td.sh script

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,11 +12,15 @@ on top of Ubuntu 23.10
 
 - Get the setup script
 
+```bash
   $ wget https://raw.githubusercontent.com/canonical/tdx/main/setup-host.sh
+```
 
 - Run the script
 
+```bash
   $ chmod a+x ./setup-host.sh && sudo ./setup-host.sh
+```
 
 - Reboot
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -42,22 +52,50 @@ or on any other Ubuntu (>= 22.04):
...
+
+```bash
+  $ https://raw.githubusercontent.com/canonical/tdx/dev-review/guest-tools/run_td.sh
+  $ chmod a+x run_td.sh
+```
+
+To run the TD:
+
+```bash
   $ TD_IMG=/tmp/tdx-guest-ubuntu-23.10.qcow2 ./run_td.sh
```
fanchen2, line 93: Do you think we can remind the users to move the created qcow2 from /tmp to a common folder? I'm not sure, this is not a TDX specific thing, the files in /tmp will be removed after reboot, just in case the users don't remember.(I forgot it myself in the beginning...)

petitfou: Ok i will do that, thanks

PR #2, opened 2023-10-30
Update README.

bktan8:


README.md
```diff
@@ -80,82 +82,74 @@ Example output:
...
-- Log into the guest
+2. Log into the guest. 
 
-The script will run the TD in the background and if the guest image has been created from scratch previously,
-it can be accessed via SSH:
+NOTE: The default credentials for a TD guest created from scratch is `root/123456`.   
 
 ```bash
-  $ ssh -p 10022 root@localhost
+ssh -p 10022 root@localhost
```
petitfou, line 146: This command to access the TD is solely valid for TD created from scratch (ssh is configured to accept ssh login under root, ..), if the user converts an existing VM into TD, access to the TD will be user specific, this is the reason why i put this boot section just after the "Create TD image from scratch", i think we can keep your proposed structure, it makes sense but we might want to point out this detail

petitfou (changes requested):

README.md
```diff
@@ -80,82 +82,74 @@ Example output:
...
 
-- Start the TD guest
+1. Boot up your guest.  
 
-If you have downloaded this repository, the script wil be in guest-tools.
+2. Clone this repo.
 
 ```bash
-  $ cd guest-tools
+git clone git@github.com:canonical/tdx.git
```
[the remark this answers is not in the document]
bktan8: I was more explicit in this case since we're inside the VM.

README.md
bktan8: Here's what I changed it to:
      NOTE: The example below uses the credentials for a TD guest created from scratch.    
      If you converted your own guest, please use your original credentials.

README.md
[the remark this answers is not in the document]
petitfou: you are right, i missed that

README.md
petitfou: it is not really a big deal here but to be precise, I meant, besides the credentials, the ssh command also is specific to TD created from scratch, more precisely, the proposed ssh user root is only applicable to the TD created from scratch, if the user uses an existing VM, he/she might want to access to the TD under a different username ...

bktan8: When I use "credentials", I am referring to both username and password.  Is that ok?

petitfou: ok, got it

petitfou (approved):
LGTM

PR #3, opened 2023-10-31
README : use https instead of ssh for git clone

petitfou:
https is simpler and has more chance to success for usages with proxy


bktan8 (approved):
LGTM

PR #6, opened 2023-11-01
Work around for login prompt missing issue

petitfou:
The released intel-opt kernel has an issue with video mode setting that made display unavailable for login prompt on the TTY, as a work around, disable the video mode setting in the kernel


bktan8 (approved):
LGTM

PR #12, opened 2023-11-13
List supported Ubuntu guest versions: 23.04 and 23.10

bktan8:


bktan8:
Not needed anymore since 23.04 is EOL on January 25, 2024.

PR #19, opened 2024-01-11
README : add instructions to build packages from source

petitfou:


fanchen2 (commented):
Tested, these steps work well, .deb packages can be generated with these steps.

Just got warning when source qemu:
W: Download is performed unsandboxed as root as file 'qemu_8.0.4+dfsg.orig.tar.xz' couldn't be accessed by user '_apt'. - pkgAcquire::Run (13: Permission denied)

And after this step, we can see 4 files, one folder for qemu, two tar.xz file and one .dsc file.
Do we need to point out which one is the source code and ask the user don't need to untar the tar.xz?

Do we need to guide the users how to install or which to install?

petitfou:
> Tested, these steps work well, .deb packages can be generated with these steps.
> 
> Just got warning when source qemu: W: Download is performed unsandboxed as root as file 'qemu_8.0.4+dfsg.orig.tar.xz' couldn't be accessed by user '_apt'. - pkgAcquire::Run (13: Permission denied)

This is a warning because you executed the instructions under root, 

> 
> And after this step, we can see 4 files, one folder for qemu, two tar.xz file and one .dsc file. Do we need to point out which one is the source code and ask the user don't need to untar the tar.xz?
> 
Yes, i can add more details to guide the users on that

> Do we need to guide the users how to install or which to install?

In the provided link for detailed instructions from community, there is a section for installing the packages, i can add the link to this section explicitly to help users

fanchen2 (approved):
LGTM

PR #23, opened 2024-01-17
Add vsock device for TD Quote Generation

ScottR-Intel:
Add vsock device to libvirt guest xml template to allow TD quote generation when a TD is launched with libvirt.


fanchen2 (approved):
Looks good to me, VM boot successfully and vsock works well with it.

PR #24, opened 2024-01-18
Add checkpoint for create-td-image.sh

fanchen2:
With different host configuration and network env, script may fail at any unexpected point, add more checkpoint for it.


petitfou (approved):
Thanks

PR #29, opened 2024-01-25
Set default path of TD image to right location

### LINKED ISSUE #28: Check image presence and change default path in run_td.sh
The script `run_td.sh` does not check if the image is present that should be used to start a TD.
The script will just silently fail if the image is not present.

Additionally, changing the default path from `ubuntu-23.10-td.qcow2` to `image/tdx-guest-ubuntu-23.10.qcow2` would match the default path if customer strictly follows the project's README. The `TD_IMG=<path_to_td_qcow2_image>` part could even be removed from `TD_IMG=<path_to_td_qcow2_image> ./run_td.sh` in the README making the start of a TD easier.

petitfou:
Enhance script output on error

Fix #28

### DIFF
--- a/guest-tools/run_td.sh
+++ b/guest-tools/run_td.sh
@@ -24,7 +24,9 @@ if [ "$1" = "clean" ]; then
     exit 0
 fi
 
-TD_IMG=${TD_IMG:-${PWD}/image/tdx-guest-ubuntu-23.10.qcow2}
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+TD_IMG=${TD_IMG:-${SCRIPT_DIR}/image/tdx-guest-ubuntu-23.10.qcow2}
 TDVF_FIRMWARE=/usr/share/ovmf/OVMF.fd
 
 if ! groups | grep -qw "kvm"; then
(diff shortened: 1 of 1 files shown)


guest-tools/run_td.sh
```diff
@@ -10,12 +17,14 @@ cleanup() {
...
 
+trap "on_exit" EXIT
+
 cleanup
 if [ "$1" = "clean" ]; then
     exit 0
 fi
 
-TD_IMG=${TD_IMG:-${PWD}/ubuntu-23.10-td.qcow2}
+TD_IMG=${TD_IMG:-${PWD}/image/tdx-guest-ubuntu-23.10.qcow2}
```
fidencio, line 27: I'd suggest to **not** do this, but rather do something like
```sh
SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```

And then get the image based on the SCRIPT_DIR, as otherwise you'll still see a failure depending from where you call the script.

BFuhry: I agree that it would even better if it would not matter from where the script is called

petitfou: the script is supposed to be used following the README, but i will do it, thanks

PR #36, opened 2024-02-20
Remove useless apt cmd and IMA config in TDX image creation

jialeif:
The TDX setup process is nice and won't break the apt system, so "apt install --fix-broken" is redundant.
Also remove IMA parameters from guest kernel cmdline to avoid confusion. Enabling TDX doesn't require kernel IMA at all.


hector-cao (approved):
LGTM

PR #43, opened 2024-03-21
Add libvirt multi-guest support

frenchwr:
Summary of important changes/features:

* Change name of script from `run_td_virsh.sh` --> `td_virsh_tool.sh` to reflect ability to not only run TDs but also clean/destroy them
* For host port forwarding pass port 0 in guest XML, which allows the kernel to allocate an available port from the host's ephemeral port range
* Create overlay qcow2 image that is unique to each TD
* Leverage libvirt's vsock `cid auto='yes'` feature to ensure a unique vsock CID for each TD

## Example Usage 

```bash
$ ./td_virsh_tool.sh
Domain td_guest-1 running with vsock CID: 3, ssh -p 44441 root@localhost

$ ssh -p 44441 root@localhost "lsb_release -a; ls /dev/tdx*"
Distributor ID:	Ubuntu
Description:	Ubuntu 23.10
Release:	23.10
Codename:	mantic
/dev/tdx_guest

$ ./td_virsh_tool.sh -n 2
Domain td_guest-2 running with vsock CID: 4, ssh -p 41595 root@localhost
Domain td_guest-3 running with vsock CID: 5, ssh -p 38111 root@localhost

$ ./td_virsh_tool.sh -c td_guest-2
Destroying domain td_guest-2.
Waiting for VM to shutdown ...

$ ./td_virsh_tool.sh -c all
Destroying domain td_guest-1.
Waiting for VM to shutdown ...
Destroying domain td_guest-3.
Waiting for VM to shutdown ...
```


guest-tools/td_virsh_tool.sh
```diff
@@ -0,0 +1,224 @@
...
+    if [ $? -ne 0 ] || [ ! -f ${xml_template_path} ]; then
+        echo "Set libvirt guest XML template path with XML_TEMPLATE"
+        exit 1
+    fi
+}
+
+set_domain() {
+    local domain_id=1
+    while $(virsh list --state-running --name |
+        grep -q "\-${domain_id}"); do
```
hector-cao, line 121: we are using a incremental counter for domain identification, i think this is hard to manage, we have to find the free id each time, using something like UUID is more robust i would think, it is kind of random number (with a very low probability of collision) when you want to generate an ID for a new domain, just generate an UUID. This will also be more robust wrt to domains that user created that are not related to TDX

hector-cao: One challenge of using UUID is however to handle nicely the clean up of qcow image

hector-cao: If we want to stick with the counter, perhaps add a DOMAIN_PREFIX in the "grep" to make it more  robust

guest-tools/td_virsh_tool.sh
```diff
@@ -0,0 +1,224 @@
...
+    domains_all=$(virsh list --all --name)
+    if echo ${domains_all} | tr " " "\n" | grep -Fxq ${domain_to_clean}; then
+        destroy ${domain_to_clean}
+    else
+        echo "Unknown domain ${domain_to_clean}"
+        echo "Existing domains: $(echo ${domains_all} | tr -d "\n")"
+        exit 1
+    fi
+    exit 0
+}
```
hector-cao, line 204: should we clean the qcow image ?

guest-tools/td_virsh_tool.sh
```diff
@@ -0,0 +1,224 @@
...
+
+clean_all() {
+    for domain_to_clean in $(virsh list --all --name | grep ${DOMAIN_PREFIX}); do
+        destroy ${domain_to_clean}
+    done
+    exit 0
+}
```
hector-cao, line 211: should we clean all the qcow images ?

hector-cao (changes requested):
Great work, some comments

guest-tools/td_virsh_tool.sh
frenchwr: I like the UUID suggestion. It also occurred to me that this will be more robust in case multiple users are using the script on the same system at once. 

One issue I see is that it looks like domains already have UUIDs assigned:

```bash
$ virsh domuuid td_guest-1
8247b776-b3c9-4be4-8994-be440b934373
```

To avoid confusion, maybe I could append a random string (making use of `/dev/urandom`) or a date string on the end of the domain name.

* random string: `td_guest-j8uw4sNkehoZMIs`
* date-based (with millisecond resolution): `td_guest-20240322:093122434`

guest-tools/td_virsh_tool.sh
frenchwr: I'm doing this [here](https://github.com/canonical/tdx/blob/42b851d24ebd9eb581efa640300b3e13e5c0fae1/guest-tools/td_virsh_tool.sh#L191).

guest-tools/td_virsh_tool.sh
hector-cao: I meant that indeed, append UUID to the domain name

frenchwr: Done. Note that I'm embedding a random string in the qcow2 overlay images for each guest to ensure uniqueness. Then I use `virsh dumpxml` to grab the qcow2 path for cleanup. 

```bash
$ ./td_virsh_tool.sh 
Domain td_guest-468ee3e2-cbc1-4c8e-ab69-51f308fa2211 running with vsock CID: 3, ssh -p 33593 root@localhost

$ ./td_virsh_tool.sh -c td_guest-468ee3e2-cbc1-4c8e-ab69-51f308fa2211
Destroying domain td_guest-468ee3e2-cbc1-4c8e-ab69-51f308fa2211.
Waiting for VM to shutdown ...
```

hector-cao (approved):
Perfect ! +1

PR #47, opened 2024-03-27
setup-tdx-host.sh : improve host setup robustness

### LINKED ISSUE #31: setup-tdx-host.sh doesn't work if it's not the first time to install it
I installed this release and remove the default 1003 kernel for some reason, but when I try to install it again, I failed to install it by executing setup-tdx-host.sh or "apt install linux-image-6.5.0-1003-intel-opt", it says everything is ok "The setup has been done successfully", "linux-image-6.5.0-1003-intel-opt is already the newest version (6.5.0-1003.3)", but no 1003 kernel found in /boot. 
And I found "apt autoremove linux-image-6.5.0-1003-intel-opt" works for me, it installed linux-image-6.5.0-1003-intel-opt successfully, I can found the kernel on /boot now.

hector-cao:
- do not use meta package kobuk-tdx-host anymore there is no value added by using meta package because we already have this script and the use of meta-package can induce some robustness problem when using already has qemu/libvirt installed in the system when doing the setup
- force kernel -intel selection for next boot

Fix #31

### DIFF
--- a/setup-tdx-host.sh
+++ b/setup-tdx-host.sh
@@ -5,6 +5,19 @@ _error() {
   exit 1
 }
 
+# grub: switch to kernel version
+grub_switch_kernel() {
+    KERNELVER=$1
+    MID=$(awk '/Advanced options for Ubuntu/{print $(NF-1)}' /boot/grub/grub.cfg | cut -d\' -f2)
+    KID=$(awk "/with Linux $KERNELVER/"'{print $(NF-1)}' /boot/grub/grub.cfg | cut -d\' -f2 | head -n1)
+    cat > /etc/default/grub.d/99-tdx-kernel.cfg <<EOF
+GRUB_DEFAULT=saved
+GRUB_SAVEDEFAULT=true
+EOF
+    grub-editenv /boot/grub/grubenv set saved_entry="${MID}>${KID}"
+    update-grub
+}
+
 apt update
 apt install --yes software-properties-common &> /dev/null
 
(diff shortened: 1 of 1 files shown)


setup-tdx-host.sh
```diff
@@ -5,6 +5,21 @@ _error() {
   exit 1
 }
 
+KERNEL_VERSION=6.5.0.1003.3
+
+# grub: switch to kernel version
+grub_switch_kernel() {
+    KERNELVER=$1
+    MID=$(awk '/Advanced options for Ubuntu/{print $(NF-1)}' /boot/grub/grub.cfg | cut -d\' -f2)
+    KID=$(awk "/with Linux $KERNELVER/"'{print $(NF-1)}' /boot/grub/grub.cfg | cut -d\' -f2 | head -n1)
```
frenchwr, line 14: On systems I've checked these lines won't match with the KERNEL_VERSION shown above. The existing grub.cfg on these systems shows:

```
menuentry 'Ubuntu, with Linux 6.5.0-1003-intel-opt'...
```

frenchwr (commented):
Left one comment but otherwise looks good!

setup-tdx-host.sh
hector-cao: The text you shown is not the version for apt

frenchwr: Right, but `awk` won't match the appropriate line from `grub.cfg` because it uses a different version string compared to the version from `apt` (KERNELVER is using the `apt` version string). Unless I'm missing something.

hector-cao: Ah, you are right, let me check

setup-tdx-host.sh
```diff
@@ -22,9 +35,22 @@ EOF
...
+apt install --yes --allow-downgrades \
+    linux-intel-opt \
+    qemu-system-x86 \
+    libvirt-daemon-system \
+    libvirt-clients \
+    ovmf \
+    tdx-tools-host
+
+# detect the intel-opt current release (example : 6.5.0-1003-intel-opt)
+KERNEL_RELEASE=$(apt show linux-image-intel-opt 2>&1 | gawk 'match($0, /Depends: linux-image-(.+)/, a) {print a[1]}')
```
frenchwr, line 47: It looks like `gawk` comes installed on Ubuntu Server and I think it's safe to assume that won't change in future releases, but if you wanted to be extra safe we could `apt install` it above.

frenchwr (approved):
:+1:

fanchen2 (approved):
Great work guys! I downgraded my QEMU to non-tdx version and removed Kobuk kernel, then tested this script and reboot, it works perfectly!

PR #50, opened 2024-04-03
Improve attestation setup script

hector-cao:
- Add reboot instruction in README  
    after the setup script run for attestation, we need to reboot
    to have the right owner & group for /dev/sgx_provision (from udev rule)

- Remove README section on use_secure_cert  
    with the version 1.18-0ubuntu2 of libsgx-dcap-default-qpl
    the use_secure_cert is set to false by default

- setup-attestation-host.sh : do not set user and group for /dev/sgx_provision  
    the udev rules will take care of it

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,6 +308,15 @@ cd tdx/attestation
 sudo ./setup-attestation-host.sh
 ```
 
+`Reboot` the system and verify that sgx devices have proper user and group.
+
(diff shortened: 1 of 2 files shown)


bktan8 (approved):
LGTM

frenchwr (approved):
Looks good!

PR #55, opened 2024-04-10
Add list command to td_virsh_tool

hector-cao:
To be able to list current VMs, this is important to see the forwarding ports for ssh connection

### DIFF
--- a/guest-tools/td_virsh_tool.sh
+++ b/guest-tools/td_virsh_tool.sh
@@ -44,7 +44,7 @@ main() {
         create_overlay_image
         create_domain_xml
         boot_vm
-        echo_ssh_cmd
+        echo_ssh_cmd ${domain}
     done
 }
 
@@ -59,6 +59,7 @@ Available options:
 -h,   --help            Print this help and exit
 -c D, --clean D         Clean domain with name D (use "all" to clean all)
 -n N, --n_instances N   Launch N instances
+-l,   --list            List VMs
 
 Environment variables:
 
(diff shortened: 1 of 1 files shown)


guest-tools/td_virsh_tool.sh
```diff
@@ -59,6 +59,7 @@ Available options:
 -h,   --help            Print this help and exit
 -c D, --clean D         Clean domain with name D (use "all" to clean all)
 -n N, --n_instances N   Launch N instances
+-l, --list              List VMs
```
frenchwr, line 62: Could you align `--list` with `--n_instances` in line above to make the formatting nicer?

guest-tools/td_virsh_tool.sh
```diff
@@ -212,6 +218,12 @@ clean_all() {
     exit 0
 }
 
+print_all() {
+    for domain_to_print in $(virsh list --all --name | grep ${DOMAIN_PREFIX}); do
```
frenchwr, line 222: I think `virsh list --state-running --name` is better here to avoid including TDs that are shut off or in some other state that makes the guest inaccessible (the guests shouldn't be in this state but just in case). Otherwise you can end up with something like:

```
$ ./td_virsh_tool.sh --list
Domain td_guest-4fb86744-a222-4174-a428-43f64c2242b7 running with vsock CID: 3, ssh -p 36113 root@localhost
error: Requested operation is not valid: domain is not running
error: Requested operation is not valid: domain is not running
Domain td_guest-06087305-06c3-4b6c-958e-563f0fc29418 running with vsock CID: , ssh -p  root@localhost
error: Requested operation is not valid: domain is not running
error: Requested operation is not valid: domain is not running
Domain td_guest-5604771f-69fa-4fe6-aaa2-d5530674abba running with vsock CID: , ssh -p  root@localhost
```

frenchwr (changes requested):

frenchwr (approved):

PR #56, opened 2024-04-11
Add intel kernel as option that can be selected at guest image creation 

hector-cao:
Use has to set the environment variable TDX_GUEST_SETUP_INTEL_KERNEL at setup-tdx-guest.sh script invocation

### DIFF
--- a/setup-tdx-guest.sh
+++ b/setup-tdx-guest.sh
@@ -20,6 +20,22 @@ EOF
     update-grub
 }
 
+# select the kernel release for next boot
+# if KERNEL_RELEASE is specified, we use it
+# if not, select the latest generic kernel available on the system
+grub_set_kernel() {
+   if [ -z "${KERNEL_RELEASE}" ]; then
+      KERNEL_RELEASE=$(ls /boot/vmlinuz-*-generic 2>&1 | \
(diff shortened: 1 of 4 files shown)


setup-tdx-guest.sh
```diff
@@ -33,20 +35,25 @@ Pin: release o=LP-PPA-kobuk-team-tdx
...
    tdx-tools-guest \
    python3-pytdxmeasure
 
-grub_switch_kernel ${KERNEL_RELEASE}
+# if a specific kernel has to be used
+# TODO : install linux-modules-extra
+if [ ! -z "${KERNEL_RELEASE}" ]; then
+  apt install --yes --allow-downgrades \
+    linux-image-unsigned-${KERNEL_RELEASE}
+  grub_switch_kernel ${KERNEL_RELEASE}
```
hector-cao, line 55: Will have to handle the case where user selects the generic kernel and the current system is using other kernel (-intel kernel from previous setup for example)

hector-cao, line 55: Will have to handle the case where user selects the generic kernel and the current system is using other kernel (-intel kernel from previous setup for example)

bktan8 (approved):
LGTM

PR #57, opened 2024-04-11
Fix use of set -e option

hector-cao:
set -e will make the script stop at the check of grub kernal commandline

### DIFF
--- a/setup-tdx-host.sh
+++ b/setup-tdx-host.sh
@@ -23,6 +23,32 @@ EOF
     update-grub
 }
 
+grub_cmdline_kvm() {
+  # update cmdline to add tdx=1 to kvm_intel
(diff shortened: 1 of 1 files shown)


setup-tdx-host.sh
```diff
@@ -23,6 +23,32 @@ EOF
     update-grub
 }
 
+grub_cmdline_kvm() {
+  # update cmdline to add tdx=1 to kvm_intel
+  grep -E "GRUB_CMDLINE_LINUX.*=.*\".*kvm_intel.tdx( )*=1.*\"" /etc/default/grub &> /dev/null
```
frenchwr, line 28: Reading this initially I did not expect it to work (intuitively I expected the function to exit if the `grep` returned no match), but then I learned that `set -e` does not apply in certain edge cases such as this one. Bash is such a strange language. :see_no_evil:

frenchwr (approved):

PR #59, opened 2024-04-12
cleanup guest setup script

hector-cao:
The guest setup script is originally derived from Intel MVP tools For 24.04, several features are not useful and are removed The script is also clean up to be efficient


guest-tools/image/create-td-image.sh
```diff
@@ -138,16 +143,8 @@ create_guest_image() {
     else
         error "Failed to copy ${CLOUD_IMG} to /tmp"
     fi
-}
 
-config_guest_env() {
```
fanchen2, line 143: For me personally, it is useful, I don't need to setup some env variable on guest again.

hector-cao: for networking access through enterprise proxy ?

fanchen2: Yes, but I agree with you, just remove it now, LGTM.

PR #64, opened 2024-04-15
setup-tdx-guest : install modules extra for generic kernel

hector-cao:
The tdx-guest module is still in modules-extra, we have to install it for the guest if it uses the generic kernel


fanchen2 (approved):
LGTM

PR #66, opened 2024-04-18
Fix shellcheck warnings

ThibF:
Fix shellchecks warnings in setup-tdx* and in guest-tools/td_virsh_tools.sh as they can hide errors.


hector-cao (approved):
LGTM

PR #69, opened 2024-04-19
corrections and updates

fanchen2:
Reduce the vcpu number of guest to reasonable numbers
TDX does not support multi-sockets for TD so set the socket to 1
Increase timeout for virt-install when creating guest image
README : improve host verification of tdx initialization


hector-cao (approved):
LGTM

PR #74, opened 2024-04-26
change kernel installation policy

hector-cao:
so far, a specific kernel version (6.8.0-1001-intel) is installed and the system will stick to it (when setup script is run again or at system upgrade). we would like to get the latest version of the kernel (for example, when 6.8.0-1002-intel will be released) automatically without having to modify the script.

### DIFF
--- a/setup-tdx-common
+++ b/setup-tdx-common
@@ -28,3 +28,44 @@ Unattended-Upgrade::Allowed-Origins {
 Unattended-Upgrade::Allow-downgrade "true";
 EOF
 }
+
+# get the current kernel version from the kernel flavour/type
+# we are using
+# the input will be the kernel meta package name of type : linux-image-<flavour|type>
+# examples : linux-image-intel, linux-image-generic, linux-image-aws, ...
+get_kernel_version() {
+    local kernel_flavour=$1
+    local kernel_version
+    kernel_version=$(apt show ${kernel_flavour} 2>&1 | gawk 'match($0, /Depends:.* linux-image-([^, ]+)/, a) {print a[1]}')
+    echo $kernel_version
+}
+
+# grub: switch to kernel version
+grub_switch_kernel() {
+    KERNELVER=$1
+    MID=$(awk '/Advanced options for Ubuntu/{print $(NF-1)}' /boot/grub/grub.cfg | cut -d\' -f2)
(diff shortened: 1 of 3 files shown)


setup-tdx-common
```diff
@@ -28,3 +28,44 @@ Unattended-Upgrade::Allowed-Origins {
...
+
+# select the kernel release for next boot
+# NB : this function will read/write the global variable KERNEL_RELEASE
+# if KERNEL_RELEASE is specified, we use it
+# if not, select the latest generic kernel available on the system and update the KERNEL_RELEASE var
+grub_set_kernel() {
+    if [ -z "${KERNEL_RELEASE}" ]; then
+      KERNEL_RELEASE=$(find /boot/vmlinuz-*-generic 2>&1 | \
+                      /usr/lib/grub/grub-sort-version -r 2>&1 | \
+                      gawk 'match($0 , /^\/boot\/vmlinuz-(.*)/, a) {print a[1];exit}')
```
frenchwr, line 64: I believe `gawk` is installed via apt in the `setup-tdx-host.sh` script but not the `setup-td-guest.sh` script. Since `gawk` is only being used from `setup-tdx-common` (I think that's right) it's probably best to apt install it here.

setup-tdx-common
```diff
@@ -28,3 +28,44 @@ Unattended-Upgrade::Allowed-Origins {
...
+    fi
+    if [ -z "${KERNEL_RELEASE}" ]; then
```
frenchwr, line 66: If I'm reading the comment above this function correctly, this second if block should be removed, no?

frenchwr (commented):
A few comments. I'm also wondering if we should add a comment to the README that both of the `setup-*` scripts should be used for upgrading the kernel.

setup-tdx-common
hector-cao: thanks, i add the installation in the guest setup script because not all scripts that include `setup-tdx-common` will necessarily call this function

setup-tdx-common
hector-cao: i don't think so, this is the error check : if we successfully determined the kernel version

frenchwr: Ah I see, this is checking whether KERNEL_RELEASE was set correctly in the previous if block.

frenchwr (approved):
Approving, assuming you've tested. :)

hector-cao:
> Approving, assuming you've tested. :)

Thanks, yes, I've tested

Issue #68, opened 2024-04-19
Boot TDVM with td_virsh_tool.sh, write some data on VM, poweroff and use td_virsh_tool.sh to boot TDVM again, data is gone


fanchen2:
Reproduce steps:
**farrah@emr-111144:~/tdx/guest-tools$ ./td_virsh_tool.sh**
Domain td_guest-18fae13e-282b-4a39-8734-0cd500234622 running with vsock CID: 3, ssh -p 38383 root@localhost
**farrah@emr-111144:~/tdx/guest-tools$ ssh -p 38383 root@localhost**

Last login: Thu Apr 18 21:04:18 2024 from 10.0.2.2
**root@tdx-guest:~# touch mytesting0419**

root@tdx-guest:~# ls
mytesting0419  test

**root@tdx-guest:~# poweroff**

Connection to localhost closed.
**farrah@emr-111144:~/tdx/guest-tools$ ./td_virsh_tool.sh**
Domain td_guest-6b55841e-8bd5-436c-86e2-e17e9b80dedf running with vsock CID: 3, ssh -p 44575 root@localhost
farrah@emr-111144:~/tdx/guest-tools$ ssh -p 44575 root@localhost

Last login: Thu Apr 18 21:04:18 2024 from 10.0.2.2
**root@tdx-guest:~# ls**
test

fanchen2:
The workaround is use virsh to start the same VM again, and use " ./td_virsh_tool.sh --list" to get new host_forward for guest ssh.

frenchwr:
Improved docs around this issue in https://github.com/canonical/tdx/pull/82, and an improved version of the script will be offered soon.

Issue #67, opened 2024-04-19
The ssh port generated by td_virsh_tool.sh doesn't work after poweroff VM and boot again by virsh


fanchen2:
Reproduce steps:
**farrah@emr-111144:~/tdx/guest-tools$ ./td_virsh_tool.sh**
Domain td_guest-f2b48482-60e8-4498-8ed2-6a335beccab7 running with vsock CID: 3, ssh -p 45457 root@localhost
**farrah@emr-111144:~/tdx/guest-tools$ ssh -p 45457 root@localhost**
...
Last login: Thu Apr 18 21:04:18 2024 from 10.0.2.2
**root@tdx-guest:~# poweroff**

Connection to localhost closed.
**farrah@emr-111144:~/tdx/guest-tools$ virsh**
Welcome to virsh, the virtualization interactive terminal.

Type:  'help' for help with commands
       'quit' to quit

virsh # list --all
 Id   Name                                            State

 td_guest-f2b48482-60e8-4498-8ed2-6a335beccab7   shut off

**virsh # start td_guest-f2b48482-60e8-4498-8ed2-6a335beccab7**
Domain 'td_guest-f2b48482-60e8-4498-8ed2-6a335beccab7' started

virsh # quit

**farrah@emr-111144:~/tdx/guest-tools$ ssh -p 45457 root@localhost**
ssh: connect to host localhost port 45457: Connection refused
farrah@emr-111144:~/tdx/guest-tools$

fanchen2:
Per @frenchwr , we can use "./td_virsh_tool.sh --list" to get the new port.
If recommend this method, it should be added to readme.

PR #82, opened 2024-04-30
Improve td_virsh_tool docs

frenchwr:
Addresses https://github.com/canonical/tdx/issues/67 and https://github.com/canonical/tdx/issues/68

After considering how best to update the script to avoid confusion around the issues above, and through discussions with @hector-cao , I eventually decided it best to limit the changes to docs improvements. 

The script is already quite complex as it's managing overlay images, host ssh ports, and CIDs that are unique to each guest (to enable multi guest support) with its own `td_guest` naming prefix (almost like a namespace). We want to avoid making it more complex and to avoid re-implementing all of `virsh`'s functionality. Interacting with the guests through `virsh` directly is more future-proof, while hopefully the script is convenient for creating multiple guests in a reproducible way.

@fanchen2 let us know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,16 @@ for destroying and cleaning up unwanted TDs.
 ./td_virsh_tool.sh -c all
 ```
 
+To see all running guests managed by the tool, pass the `--list` flag:
+
+```bash
+./td_virsh_tool.sh --list
+Domain td_guest-123 running with vsock CID: 3, ssh -p 34825 root@localhost
+Domain td_guest-456 running with vsock CID: 4, ssh -p 33789 root@localhost
+```
+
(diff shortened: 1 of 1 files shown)


fanchen2:
@frenchwr I agree with you, the script is already quite complex. And also the readme and usage...to support multiple guests, we sacrificed the simplicity of this release, may make users confusing(if they miss one or two lines of the readme) or easy to make mistake when using.

frenchwr:
> @frenchwr I agree with you, the script is already quite complex. And also the readme and usage...to support multiple guests, we sacrificed the simplicity of this release, may make users confusing(if they miss one or two lines of the readme) or easy to make mistake when using.

@fanchen2 Yes, I agree. @hector-cao is working on an improved version of the script that is just a thin wrapper around `virsh`, and therefore interoperates with `virsh` in a much more intuitive way.

@hector-cao I would suggest we go ahead and merge this PR for posterity. In case a user for some reason wants to use `td_virsh_tool.sh` the corresponding README will capture the tool's usage more fully.

hector-cao (approved):
LGTM

PR #84, opened 2024-04-30
run_td.sh: add hpet=off

fanchen2:
Recommend to use tsc as the clocksource in TD guest and not to use hpet. Without hpet=off, the default clocksource in TD guest is also tsc, but the users can change the closcksouce by themselves. So we disable hpet in qemu para to ensure hpet will not be used in TD guest.

### DIFF
--- a/guest-tools/run_td.sh
+++ b/guest-tools/run_td.sh
@@ -36,7 +36,7 @@ qemu-system-x86_64 -D /tmp/tdx-guest-td.log \
 		   -name ${PROCESS_NAME},process=${PROCESS_NAME},debug-threads=on \
 		   -cpu host \
 		   -object tdx-guest,id=tdx \
-		   -machine q35,kernel_irqchip=split,confidential-guest-support=tdx \
+		   -machine q35,kernel_irqchip=split,confidential-guest-support=tdx,hpet=off \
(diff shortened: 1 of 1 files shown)


guest-tools/run_td.sh
```diff
@@ -36,7 +36,7 @@ qemu-system-x86_64 -D /tmp/tdx-guest-td.log \
 		   -name ${PROCESS_NAME},process=${PROCESS_NAME},debug-threads=on \
 		   -cpu host \
 		   -object tdx-guest,id=tdx \
-		   -machine q35,kernel_irqchip=split,confidential-guest-support=tdx \
+		   -machine q35,kernel_irqchip=split,confidential-guest-support=tdx,hpet=off \
```
hector-cao, line 39: could you confirm that in the td_guest.xml.template we already have this option off :
```
  <clock offset='utc'>
    <timer name='hpet' present='no'/>
  </clock
  
```

hector-cao (commented):
LGTM

guest-tools/run_td.sh
fanchen2: Yes, it's already off in  td_guest.xml.template

hector-cao (approved):

PR #85, opened 2024-04-30
re-design of tdvirsh

hector-cao:
rename the rool to tdvirsh to be more close to virsh
store data in /var/tmp/tdvirsh
make tdvirsh a wrapper around virsh
remove creation of multiple TDs at one call

Sample sample usage:

```
$ ./tdvirsh new
Create and run new TD ...
Id   Name                                            State 
--------------------------------------------------------------- 
1    td_guest-03350e82-e30e-4dac-b590-b6f937eddd8b   running (ssh:35869, cid:3)
```

```
$ ./tdvirsh list --all
Id   Name                                            State 
--------------------------------------------------------------- 
1    td_guest-03350e82-e30e-4dac-b590-b6f937eddd8b   running (ssh:35869, cid:3)
```

```
$ ./tdvirsh console td_guest-03350e82-e30e-4dac-b590-b6f937eddd8b
Connected to domain 'td_guest-03350e82-e30e-4dac-b590-b6f937eddd8b'
Escape character is ^] (Ctrl + ])
```


guest-tools/tdvirsh
```diff
@@ -0,0 +1,186 @@
...
+        echo "$line $extra_info"
+    done < <(virsh "$@")
+}
+
+run_td() {
+    echo "Create and run new TD ..."
+    create_overlay_image
+    create_domain_xml
+    boot_vm
+    print_all list
```
frenchwr, line 148: I think it's better to only print the line for the VM that was just booted here rather than all of the VMs.

guest-tools/tdvirsh
```diff
@@ -0,0 +1,186 @@
...
+        destroy ${domain_to_clean}
+    done
+    rm -rf ${WORKDIR_PATH}
+}
+
+print_all() {
+    while read -r line
+    do
+        local extra_info=""
+        local td_domain=$(echo $line | grep -oP " ${DOMAIN_PREFIX}-[^ ]+")
```
frenchwr, line 123: I think you can safely drop the single space in front of ${DOMAIN_PREFIX}. When I was testing I tried:

```
./tdvirsh list --name
```

and it didn't return the ssh and cid information because with the `--name` flag there is no leading space in front of the domain name.

guest-tools/tdvirsh
```diff
@@ -0,0 +1,186 @@
...
+
+    echo "Destroying domain ${domain_to_destroy}"
+
+    virsh shutdown ${domain_to_destroy} &>/dev/null
+    virsh shutdown --domain ${domain_to_destroy} &>/dev/null
+
+    echo "Waiting for VM to shutdown ..."
+    sleep 5
+
+    virsh destroy ${domain_to_destroy} &>/dev/null
```
frenchwr, line 105: I think you mentioned this during our call, `virsh shutdown` works with an ID but `virsh destroy` fails, leaving the VM behind in a powered off state. I could see a user trying to delete by ID, and in that case the output is quite misleading.

A few options we could try:

1. Validate the domain name before passing to destroy()
2. If we like the convenience of cleaning by ID, there is a `virsh domname` subcommand that converts the ID to the domain name.

frenchwr (commented):
I left a lot of comments because I enjoyed reading through it and testing. :) This is a huge improvement in terms of both usability and maintainability. :100:

guest-tools/tdvirsh
hector-cao: right, i did not catch that, even if we remove the space, it will not work if user issues : `./tdvirsh list -id`

guest-tools/tdvirsh
hector-cao: I propose the check the argument to be a domain name
```
        delete)                                                                                               
            domain_to_clean="${2-}"                                                                           
            if [[ ! "${domain_to_clean}" =~ ^${DOMAIN_PREFIX}-[^\ ]+$ ]]; then                                
                usage                                                                                         
                exit 0                                                                                        
            fi 
```

guest-tools/tdvirsh
hector-cao: i propose to override the `list` command, just ignore all the arguments to `list` of `virsh`

guest-tools/tdvirsh
```diff
@@ -0,0 +1,195 @@
...
+# Global variables
+#
+overlay_image_path=${WORKDIR_PATH}/overlay.${domain}.qcow2
+base_img_path=""
+xml_template_path=""
+created_domain=""
+
+usage() {
+    cat <<EOM
+Usage: ./$(basename "${BASH_SOURCE[0]}") [-h] [delete D]
```
frenchwr, line 25: This is missing `[new]` and `[list]`

guest-tools/tdvirsh
```diff
@@ -0,0 +1,195 @@
...
+            # do not handle other arguments allowed by virsh
+            print_all "list --all"
+            exit 0
+            ;;
+        delete)
+            domain_to_clean="${2-}"
+            if [ "${domain_to_clean}" == "all" ]; then
+                clean_all
+            else
+                if [[ ! "${domain_to_clean}" =~ ^${DOMAIN_PREFIX}-[^\ ]+$ ]]; then
```
frenchwr, line 177: This fails if the UUID is off by a digit (which could happen from a bad copy-paste) and then shows some misleading output from `destroy()`:

```
$ ./tdvirsh list
Id   Name                                            State 
--------------------------------------------------------------- 
2    td_guest-6d7f26a1-168e-4813-aa26-5127eeec388d   running (ssh:33827, cid:4)
 
$ ./tdvirsh delete td_guest-6d7f26a1-168e-4813-aa26-5127eeec388
error: failed to get domain 'td_guest-6d7f26a1-168e-4813-aa26-5127eeec388'
Destroying domain td_guest-6d7f26a1-168e-4813-aa26-5127eeec388
Waiting for VM to shutdown ...
```

We could use `set -e` at the top of the `destroy()` function so it exits on the first command using an invalid domain name.

guest-tools/tdvirsh
```diff
@@ -0,0 +1,195 @@
...
+}
+
+clean_all() {
+    for domain_to_clean in $(virsh list --all --name | grep ${DOMAIN_PREFIX}); do
+        destroy ${domain_to_clean}
+    done
+    rm -rf ${WORKDIR_PATH}
+}
+
+print_all() {
```
frenchwr, line 122: I'm trying to think of a clearer name for this function, maybe `print_virsh_wrapped()` or `print_with_guestinfo()`? I think the `print_all dominfo "${created_domain}"` below is not easy to understand because of the function name.

hector-cao: I propose to not use print_all dominfo anymore so that we do not have to rethink the function name

guest-tools/tdvirsh
hector-cao: Adding `set -e` will give us other issues like we will stop to clean up the domain and its associated data if the domain is already shutdown or destroyed ... i propose to to sanity check to make sure the domain existsn if it does, we will do all the operations no matter the outcome of each individual operation

frenchwr (approved):
Great work! :rocket:

PR #86, opened 2024-04-30
Update Readme and effected scripts

bktan8:
* Update readme
* Remove attestation setup scripts from setup-tdx-host.sh and setup-tdx-guest.sh
* Rename attestation setup-host and setup-guest.sh to setup-attestation-host.sh and setup-attestation-guest.sh to be more explicit.


bktan8:
Please ignore.  Bad PR.

PR #87, opened 2024-04-30
Update Readme and affected setup scripts

bktan8:
* Update Readme intro to include blurb about attestation and minor fixes.  
* Remove setup attestation scripts from setup-tdx-host.sh and setup-tdx-guest.sh so user will have to manually do it. 
* Rename attestation setup-host.sh and setup-guest.sh scripts to setup-attestation-host.sh and setup-attestation-guest.sh


mythi:
#89 could be fixed here

setup-tdx-guest.sh
```diff
@@ -51,6 +51,3 @@ grub_set_kernel
 if [[ "$KERNEL_RELEASE" == *-generic ]]; then
   apt install --yes "linux-modules-extra-${KERNEL_RELEASE}"
 fi
-
-# setup attestation
-"${SCRIPT_DIR}"/attestation/setup-guest.sh
```
hector-cao, line 56: I would not remove the call to the attestation script, i would make it conditional instead, the default behavior can be what you want (no setup of attestation) but if user want, they can have enable attestation setup just by setting a variable, it will also make the README shorter

hector-cao (changes requested):
Thanks

setup-tdx-guest.sh
bktan8: Regarding making the README shorter by adding the auto install via the setup-tdx-host.sh script: there's really not that much savings.  Right now (manual way), it adds 6 lines of text.  If i have to tell them to create an environment variable, etc. I think it will be almost the same amount of text to explain.  

And could be more prone to problems.  For example, If they missed (didn't read carefully, as users often do) the instruction to create the environment variable to add DCAP as part of the setup-host-tdx.sh script and when they get to step 8 to setup attestation, they don't have DCAP.  how would they know to install DCAP? I'll have to add extra instruction that installing DCAP is necessary before proceeding with the configuration and if you didn't install DCAP, here are the steps, etc.  

IMO, documenting exact steps to install DCAP is less error prone. 

What do you think?  :)

hector-cao: If they missed, they can re-run the script again, from my perspective, i would like to have a way to have a full setup at once

bktan8: ok. Made changes per your preference.

guest-tools/image/create-td-image.sh
```diff
@@ -219,6 +219,7 @@ setup_guest_image() {
     # export environment variables to guest
     # all environment variables with prefix : TDX_GUEST_SETUP_
     declare -px | grep TDX_GUEST_SETUP_ > ${CURR_DIR}/tdx-guest-setup-env
+    declare -px | grep INSTALL_ATTESTATION >> ${CURR_DIR}/tdx-guest-setup-env
```
hector-cao, line 222: Thanks @bktan8 , it is great, We already have a mechanism to pass all environment variables with name TDX_GUEST_SETUP_* , I propose we use this channel to pass the attestation installation configuration, see my proposal here : https://github.com/bktan8/tdx/pull/1

hector-cao (changes requested):

hector-cao (approved):
Thanks @bktan8,
I will approve the MR, there might be some enhancements we can do afterwards:

- Do not have to set TDX_SETUP_ATTESTATION at various scripts call (in the README), the default behavior will be "not install attestation components"
- Since we add more and more options, the way we pass variables to scripts leads to tedious command instructions, i will try to work on using an other way for a simpler README

PR #104, opened 2024-05-10
improvement setup configuration management

hector-cao:
put all variables in a config file : setup-tdx-config
clean up setup scripts in the guest at the end
allow downgrade for attestation packages


bktan8 (approved):
LGTM

PR #120, opened 2024-05-23
Fix readme for switching to -intel kernel

hector-cao:
Recently, we switched to use a configuration file to centralize the configuration for the setup process. We do not ask users to set directly env variables anymore.


frenchwr (approved):
:+1:

fanchen2 (approved):
LGTM

PR #121, opened 2024-05-23
Add configuration variable to allow to switch TDX PPA

hector-cao:
The PPA used for tdx packages is tdx-release.
For development purpose, we might want to switch back to the dev PPA (tdx). Add a variable in the config file for that.

### DIFF
--- a/setup-tdx-guest.sh
+++ b/setup-tdx-guest.sh
@@ -24,7 +24,7 @@ apt install --yes software-properties-common gawk &> /dev/null
 rm -f /etc/apt/preferences.d/kobuk-team-tdx-*
 rm -f /etc/apt/apt.conf.d/99unattended-upgrades-kobuk
 
-add_kobuk_ppa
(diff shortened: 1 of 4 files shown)


setup-tdx-guest.sh
```diff
@@ -24,7 +24,7 @@ apt install --yes software-properties-common gawk &> /dev/null
 rm -f /etc/apt/preferences.d/kobuk-team-tdx-*
 rm -f /etc/apt/apt.conf.d/99unattended-upgrades-kobuk
 
-add_kobuk_ppa
+add_kobuk_ppa ${TDX_PPA:-tdx}
```
frenchwr, line 27: Should the default here be `tdx-release`, since that's mentioned as the default in the configuration file? I guess the scenario where TDX_PPA is not defined should not happen, but if we're using a fallback here it seems like `tdx-release` would be more consistent.

hector-cao: yes, you are right, thanks

PR #138, opened 2024-06-07
Adding all pytests to checkbox list for PEK-710

spmcmillan:
All tests available were added.  Put reports generated into /var/tmp/tdxtest/reports (as checkbox calls pytest).

Removed td-perf-benchmark as it doesn't seem to complete after several hours.  Will resolve later.

This is a continuation of other PEK-710 pull request.

### DIFF
--- a/tests/checkbox/checkbox-provider-tdx/units/basic/jobs.pxu
+++ b/tests/checkbox/checkbox-provider-tdx/units/basic/jobs.pxu
@@ -83,6 +83,18 @@ command:
   export LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu/:$LD_LIBRARY_PATH
   setup-env-and-run test_boot_multiple_vms.py
 
+id: tdx-basic/td-stress-boot
+category_id: tdx-basic
+flags: simple
+_summary: Test boot stress
+depends:
+after:
+requires:
+  executable.name == 'qemu-system-x86_64'
+command:
+  export LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu/:$LD_LIBRARY_PATH
+  setup-env-and-run test_stress_boot.py
+
 id: tdx-basic/td-measurement
 category_id: tdx-basic
 flags: simple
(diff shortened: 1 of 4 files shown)


tests/checkbox/checkbox-provider-tdx/units/basic/jobs.pxu
```diff
@@ -83,6 +83,18 @@ command:
...
   setup-env-and-run test_boot_multiple_vms.py
 
+id: tdx-basic/td-stress-boot
+category_id: tdx-basic
+flags: simple
+_summary: Test boot stress
+depends:
+after:
+requires:
+  executable.name == 'qemu-system-x86_64'
```
frenchwr, line 93: Just curious how this will handle our patched, tdx-enabled QEMU? I guess since checkbox runs as a snap that it's up to us to build a snap with the custom QEMU?

tests/checkbox/checkbox-provider-tdx/units/basic/jobs.pxu
```diff
@@ -83,6 +83,18 @@ command:
...
+command:
+  export LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu/:$LD_LIBRARY_PATH
```
frenchwr, line 95: Is there any way in checkbox to define this command at the unit scope (or wherever) so you don't need to run it for each test?

frenchwr (approved):
Went ahead and approved but left a few questions for my own edification :)

tests/checkbox/checkbox-provider-tdx/units/basic/jobs.pxu
hector-cao: the tests are run against a host system components (here : kernel, qemu), for this kind of testing, no need to embed qemu inside the snap

tests/checkbox/checkbox-provider-tdx/units/basic/jobs.pxu
hector-cao: This is something that already exists before this PR, we can discuss about this for sure but not in the scope of this MR, but however, here is my comment: for this, AFAIK, i think there is only 2 scopes : global and test. No possibility to have a globe that covers category or family/set of tests.

tests/checkbox/checkbox-provider-tdx/units/basic/jobs.pxu
frenchwr: ok, so the tests depend on the tdx-enabled versions being installed on the host system?

PR #139, opened 2024-06-07
PEK-710 - fix snap build

spmcmillan:
Simple fix for the snap builds.  More work in future pulse to deal with where the guest qcow2 image gets referenced.


frenchwr (approved):

hector-cao (approved):
LGTM

PR #140, opened 2024-06-11
fix issue with test_guest_report as well as using TDX_PPA config vari…

spmcmillan:
As I was verifying another issue, I:
* Added the use of TDX_PPA config variable (from setup-tdx-config) in setup-attestation-* scripts
* Fixed issue with test_guest_report.py.


spmcmillan:
I linked this to the GH issue: https://github.com/canonical/tdx/issues/141

hector-cao (approved):
LGTM

PR #147, opened 2024-06-17
Update TDX test paths, mention hours as test duration, add exemplary …

OrestTa:
…report.xml, for INTEL(R) XEON(R) PLATINUM 8570 (Intel SDP)

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -20,27 +20,62 @@ Go to the `tests` folder.
 - Run sanity tests to check the host setup:
 
 ```
-$ sudo ./run.sh pytest bin/test_host_*.py
+$ sudo ./run.sh pytest pytest/test_host_*.py
 ```
 
 - Run sanity tests to check the guest boot:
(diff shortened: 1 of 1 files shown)


hector-cao (changes requested):
Thanks @OrestTa ! I m just wondering if we need to add an example of the test report, i dont think it adds a lot of value to users here, it makes the README longer (will shift the following sections) and will need to be maintained if we update the tests

OrestTa:
> Thanks @OrestTa ! I m just wondering if we need to add an example of the test report, i dont think it adds a lot of value to users here, it makes the README longer (will shift the following sections) and will need to be maintained if we update the tests

Hm yeah, I see, feel free to move or remove it. I think one helpful piece of information for me would have been to mention the order of magnitude for how long the tests would be taking (as there's little CLI output) so that a dev can plan this better and then understand if things are on track, or actually got stuck.

PR #152, opened 2024-06-27
licence: fix shebangs and licence

ThibF:
Shebangx needs to be on the first line of the file to be effective. Move back all license notice after the shebang.


spmcmillan (approved):
LGTM

PR #157, opened 2024-07-03
Enable secure boot for TD VM

hector-cao:
PEK-755

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,6 @@
 * [8. Setup Remote Attestation on Host OS and inside TD](#setup-remote-attestation)
 * [9. Build Packages From Source](#build-packages-from-source)
 * [10. Run Tests](#sanity-functional-tests)
-* [11. Secure boot](#secure-boot)
 
 <!-- headings -->
 <a id="introduction"></a>
@@ -233,6 +232,14 @@ A different qcow2 image (e.g., one with an `intel` kernel) can be used by settin
 	```
 	./tdvirsh -h
 	```
+
+### 6.3 Secure Boot TD
+
+To run a TD VM with secure boot enabled, use the `tdvirsh` tool with
+dedicated libvirt configuration:
+
+	cd tdx/guest-tools
+	XML_TEMPLATE=trust_domain-sb.xml.template ./tdvirsh new
 
 
 <a id="verify-td"></a>
@@ -583,12 +590,3 @@ Here are example instructions for building QEMU (for normal user with sudo right
 ## 10. Running Tests
 
 Please follow [tests/README](tests/README.md) to run Intel TDX tests.
-
-<a id="secure-boot"></a>
-## 11. Secure boot
-
-To run a TD VM with secure boot enabled, use the `tdvirsh` tool with
-dedicated libvirt configuration:
-
-	cd tdx/guest-tools
-	XML_TEMPLATE=trust_domain-sb.xml.template ./tdvirsh new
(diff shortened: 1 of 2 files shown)


fanchen2 (commented):
Tested, secure boot enabled successfully.
root@tdx-guest:~# dmesg|grep -i "secure boot"
[    0.000000] secureboot: Secure boot enabled
[    0.000000] Kernel is locked down from EFI Secure Boot mode; see man kernel_lockdown.7
[    0.012374] secureboot: Secure boot enabled

README.md
```diff
@@ -11,6 +11,7 @@
 * [8. Setup Remote Attestation on Host OS and inside TD](#setup-remote-attestation)
 * [9. Build Packages From Source](#build-packages-from-source)
 * [10. Run Tests](#sanity-functional-tests)
+* [11. Secure boot](#secure-boot)
```
bktan8, line 14: 1) Change title to "Secure Boot TD"
2) I think this can be a subsection of "6. Boot TD" since it's booting TD related.  So it would be 6.3.  If you think it deserves a section of its own, that I would make it section 7 and move everything else down.  I think making after the "10. Running Tests" doesn't give it the importance that it deserves.

frenchwr:
@hector-cao I see from:

```diff
$ diff trust_domain.xml.template trust_domain-sb.xml.template 
11c11
<     <loader>/usr/share/qemu/OVMF.fd</loader>
---
>     <loader>/usr/share/ovmf/OVMF.tdx.fd</loader>
```

that we're just pointing to a different loader in the new template. I think it would be more maintainable to instead update `tdvirsh` to accept a new flag for enabling secure boot (which would then take care of updating the loader line in the template). That way we don't need to maintain two separate template files. What do you think?

About `/usr/share/ovmf/OVMF.tdx.fd` - does that get installed by a package we maintain in our PPA? What happens if the file is not present?

hector-cao:
> @hector-cao I see from:
> 
> ```diff
> $ diff trust_domain.xml.template trust_domain-sb.xml.template 
> 11c11
> <     <loader>/usr/share/qemu/OVMF.fd</loader>
> ---
> >     <loader>/usr/share/ovmf/OVMF.tdx.fd</loader>
> ```
> 
> that we're just pointing to a different loader in the new template. I think it would be more maintainable to instead update `tdvirsh` to accept a new flag for enabling secure boot (which would then take care of updating the loader line in the template). That way we don't need to maintain two separate template files. What do you think?
>

Agree; let me see what i can do
 
> About `/usr/share/ovmf/OVMF.tdx.fd` - does that get installed by a package we maintain in our PPA? What happens if the file is not present?

It is a file of OVMF package (EDK2)
If the file is missing, the boot will fail

frenchwr:
> It is a file of OVMF package (EDK2)

Ok - I assume we're patched the package in our PPA, right? We should document the need for this (the limitations around using `-bios` in qemu, etc) some place.

> If the file is missing, the boot will fail

If you are updating `tdvirsh` to accept a new flag I would suggest checking for the file before attempting to launch the TD.

hector-cao:
> > It is a file of OVMF package (EDK2)
> 
> Ok - I assume we're patched the package in our PPA, right? We should document the need for this (the limitations around using `-bios` in qemu, etc) some place.
> 
> > If the file is missing, the boot will fail
> 
> If you are updating `tdvirsh` to accept a new flag I would suggest checking for the file before attempting to launch the TD.

I think about modifying the `tdvirsh` by adding the secure boot option but i m not convinced by the relevancy of this. I don't want to make the tool more complex for something we can handle in the machine definition file; it is more expensive to maintain source code than configuration file

hector-cao:
> places

`tdvirsh` is only a tool to show users how to create  a TD and to help them to do it quickly for testing purpose, per consequence, the configuration files are only informational and we are not required to keep the two files in sync if not necessary

frenchwr (approved):

bktan8 (approved):
LGTM

PR #160, opened 2024-07-09
Pek 743 create github action for snap build on repo push

spmcmillan:
Adding github action to create snap build on repo push using github hosted runner. Notes:

-  When using snapcraft, the created container, during snapcraft build, doesn't appear to get an IP address and, as a result, doesn't work properly without using "--destructive" option of snapcraft
- When using "--destructive" mode of snapcraft, we can't use 22.04 (jammy) since the snap is meant to be built with 24.04 (noble).  Therefore, we have to use "24.04" (noble) as the "runs-on" option for github action

### DIFF
--- a/.github/workflows/create-snap-action.yml
+++ b/.github/workflows/create-snap-action.yml
@@ -19,7 +19,7 @@ jobs:
           cd tests
           # run snapcraft in destructive mode 
           # -- network doesn't work properly otherwise
-          sudo snapcraft --destructive-mode
+          sudo snapcraft
       - name: Output logs step
         if: always()
         run: |
--- a/.github/workflows/create-snap-action.yml
+++ b/.github/workflows/create-snap-action.yml
@@ -3,17 +3,15 @@ run-name: ${{ github.actor }} is creating testing snap
 on: [push]
 jobs:
   Create-Snap-Action:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-22.04
     continue-on-error: true
     steps:
       - name: Check out repository code step
         uses: actions/checkout@v4
       - name: Installation step
         run: |
-          sudo apt update
-          sudo apt install snapcraft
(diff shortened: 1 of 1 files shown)


.github/workflows/create-snap-action.yml
```diff
@@ -0,0 +1,30 @@
+name: Do a snap build
+
+run-name: ${{ github.actor }} is creating testing snap
+
+on:
+  push:
+    branches: 
+      - main
+      - pek_743_create_snap_action
```
frenchwr, line 9: Are you going to remove this prior to merging?

frenchwr (commented):
LGTM! One small comment to confirm you were only adding the feature branch temporarily for testing.

frenchwr:
I also just noticed your comments about jammy and noble - do those comments still apply since your changes this morning? It looks like we're only building for jammy now but shouldn't we be building for noble since we're merging this into the noble-24.04 branch?

hector-cao:
> I also just noticed your comments about jammy and noble - do those comments still apply since your changes this morning? It looks like we're only building for jammy now but shouldn't we be building for noble since we're merging this into the noble-24.04 branch?

i think we are building the snap with core24 (noble) and we are building it on 22.04

hector-cao (approved):
LGTM

spmcmillan:
> I also just noticed your comments about jammy and noble - do those comments still apply since your changes this morning? It looks like we're only building for jammy now but shouldn't we be building for noble since we're merging this into the noble-24.04 branch?

They no longer apply.  The network issues I was seeing with the container are dealt with by the action.  (https://github.com/canonical/tdx/actions/runs/9861054202/job/27228650775).  Apparently docker related packages were interfering with networking on the containers used by snapcraft (the action does exactly what I did, except it adds a line for /usr/bin/sudo iptables -P FORWARD ACCEPT to deal with that issue).

PR #163, opened 2024-07-15
Enhance tdvirsh new to accept various parameters

BFuhry:
- switch from environment vars to command parameters
- add option to define the main memory size
- add option to define amount of vcpus


hector-cao:
Hello @BFuhry ; the `tdvirsh` script allows you already to specify a customized template file; in that customized template file, you can do all sorts of customization; don't you think that this can suit your needs instead of making `tdvirsh` more complex and more  difficultt to maintain ? cc @frenchwr

BFuhry:
Ok, see PR #167

PR #170, opened 2024-07-22
trap on errors

matti:
Most of the shell scripts in this repo are missing `set -e` to stop on error.

For example, if `apt update` fails due to network error the script will just continue ( https://github.com/canonical/tdx/blob/8e2fced50bccd89275b3acf1b644dd915c4d7224/guest-tools/image/setup.sh#L7)

This PR adds not only that but also -E to stop inside of functions and does not just exit, but reports the error location nicely.

Next step would be to add this to all shell scripts.

### DIFF
--- a/guest-tools/image/setup.sh
+++ b/guest-tools/image/setup.sh
@@ -4,6 +4,20 @@
 #
 # This file is part of tdx repo. See LICENSE file for license information.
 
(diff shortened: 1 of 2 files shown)


hector-cao (approved):
LGTM; thanks !

PR #171, opened 2024-07-24
Dev improve log

### LINKED ISSUE #93: Show package installation status at the beginning of create-td-image.sh
During the first use of create-td-image.sh, it installs a bunch of packages.  The status is being captured in /tmp/logfile.  It would be good to show what's happening on the screen as well so it doesn't give the impression it's hanging, causing the user to wonder and maybe press Ctl-C.

### LINKED ISSUE #112: make scripts more verbose
It would be good that script would say what it's currently doing, it looks like a hang:

image

hector-cao:
Improve create-td-image script log to invite users to wait for long running operations

Fix #93
Fix #112


matti:
image

ThibF (approved):
LGTM

PR #176, opened 2024-07-25
Remove redundant -cpu option of qemu

kajinamit:
The domain definition already enforces usage of host-passthrough CPU mode, so adding -cpu host is redundant.


hector-cao (approved):
LGTM; thanks

PR #177, opened 2024-07-25
Add support to use development ppas

ThibF:


setup-tdx-host.sh
```diff
@@ -90,7 +118,12 @@ rm -f /etc/apt/apt.conf.d/99unattended-upgrades-kobuk
 # stop at error
 set -e
 
-add_kobuk_ppa ${TDX_PPA:-tdx-release}
+if [ "${dev:-0}" -eq "1" ]; then
+  add_kobuk_ppa "tdx-proposed"
```
hector-cao, line 122: @ThibF We already can tune the TDX_PPA by modifying the setup config file : https://github.com/canonical/tdx/blob/5e4a70877028bd73fa02b218eb5a502f96d05f6e/setup-tdx-config#L26

I m wondering if we have to add an argument to the script or take the advantage of the current configuration mechanism to achieve this need; 

We could make the MR to accept the `TDX_PPA` field to accept a list of PPAs

```
TDX_PPA=tdx tdx-proposed
```

setup-tdx-common
```diff
@@ -26,24 +26,25 @@
 # in case the packages in the PPA have older versions than the ones
 # from Ubuntu archive
 add_kobuk_ppa() {
```
hector-cao, line 28: Could we create an other function add_kobuk_ppas() that will call add_kobuk_ppa ?

setup-tdx-host.sh
ThibF: I would prefer both as command line help is a also natural way to discover script capabilities.
I added also a line about the configuration to make it more visible.
I updated my PR to support a list of PPA, but made it compatible that you can either use the flag or the config file.

setup-tdx-guest.sh
```diff
@@ -47,13 +47,45 @@ fi
 
 source ${SCRIPT_DIR}/setup-tdx-common
 
+usage()
```
hector-cao, line 50: Could we replace the adding of an argument to this script by:
- 1) letting user modify the config file (modify `TDX_PPA` variable) to switch to dev mode.
- 2) providing a custom config file for dev mode and allow the script to accept a custom configuration file

Personally, i would go for option 1; the effort to modify `TDX_PPA` is not big and that can leave some freedom to developer to add more PPAs

What do you think ?

setup-tdx-host.sh
hector-cao: Sure; i would think that it is a good point to not add more required feature to the scripts; that allows to lower the maintenance cost and futhermore, the conf approach is more flexible. But ok, let me think about it

hector-cao: If we want the script argument, we will need to add the capability that allow  `setup-tdx-guest.sh` being called in `dev` mode when we create an image, this is the principal use case of `setup-tdx-guest.sh`

setup-tdx-common
ThibF: Done

ThibF:
@hector-cao Following our last discussion, I updated the PR to rely on the config file

hector-cao (approved):
LGTM; thanks !

PR #190, opened 2024-08-13
Fix link in issue templates

frenchwr:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -17,5 +17,4 @@ Steps to reproduce the behavior:
 A clear and concise description of what you expected to happen.
 
 **System report**
-Please run the [system-report.sh](../../system-report.sh) script on your
-host system and copy the output below.
+Please run the [system-report.sh](https://github.com/canonical/tdx/blob/f58ad41/system-report.sh) script on your host system and copy the output below.
```
hector-cao, line 20: I dont think that is necessary to put a absolute url link to the script here, the user will use the local script they have as part of the tdx repo . Relative path shoud be enough to instruct people how to find the script locally. What do you think ?

hector-cao (changes requested):

.github/ISSUE_TEMPLATE/bug_report.md
frenchwr: Unfortunately relative paths don't work from an issue, so I think you're stuck with a permalink. The other option is to not link the script at all and just say something like:

Please run the `system-report.sh` script (located in the root directory of this repo) on your host system and copy the output below.

Do you prefer one over the other? I think either is fine.

hector-cao: When i mean relative link; i did not mean a link but a path so i will go for the second option

hector-cao (approved):
LGTM

PR #193, opened 2024-08-14
system-report.sh : various fixes

### LINKED ISSUE #192: Improve the system-report.sh
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
I very much like the `system-report.sh` script that was added to the repo.

I have a few related suggestions:
1. While executing `system-report.sh` on a fresh Ubuntu installation, the “set_msr_result_string” function leads to multiple `rdmsr: command not found` entries, because the `msr-tools` package is not installed by default. You might want to check the presence of the command at the beginning of the script and exit if not present.
2. For the `set_msr_result_string` function in `system-report.sh`, you also might want to look at https://cc-enabling.trustedservices.intel.com/intel-tdx-enabling-guide/05/host_os_setup/#check-intel-tdx-enablement at which we provide details about the MSRs you are reading. I especially consider “HW_ENCRYPT_ENABLE” a bit misleading. For the number of key, there is also no expectation of having a number > 32. 
3. You might consider to integrate the `check-production.sh` script into `system-report.sh`. This would make `system-report.sh` more standalone. Currently, `check-production.sh` is only used in step 8.1, but the following code change would be enough to replace the command in step 8.1 with  `./system-report.sh | grep -i 'production status'`.

    ```
    diff --git a/attestation/check-production.sh b/attestation/check-production.sh
    index 04f1e16..86e7529 100755
    --- a/attestation/check-production.sh
    +++ b/attestation/check-production.sh
    @@ -47,7 +47,7 @@ fi
    
    echo "CPU: ${CPU_GEN}"
    if [ "${PROD}" = "0" ]; then
    -    echo "Production"
    +    echo "Production status: production"
    else
    -    echo "Pre-production"
    +    echo "Production status: pre-production"
    fi
    ```
4. The file `/var/log/mpa_registration.log` is not present if the corresponding package is not installed. As a result, the script currently returns the error `tail: cannot open '/var/log/mpa_registration.log' for reading: No such file or directory` instead of just not not returning the log. I'd propose the following change:

    ```
    diff --git a/system-report.sh b/system-report.sh
    index aa97629..88d5a8e 100755
    --- a/system-report.sh
    +++ b/system-report.sh
    @@ -122,7 +122,7 @@ print_section "QGSD service status" "${result}"
     result=$(systemctl status pccs 2>&1)
     print_section "PCCS service status" "${result}"
    
    -result=$(tail -n 30 /var/log/mpa_registration.log)
    +result=$(tail -n 30 /var/log/mpa_registration.log 2>/dev/null)
     print_section "MPA registration logs (last 30 lines)" "${result}"
    
     printf "<======== COPY ABOVE HERE ========>\n"
    ```

**Additional context**
N/A

**System report**
N/A

hector-cao:
- Install msr-tools and load msr driver if not present this is required to read MSR values
- MSR value : rename HW_ENCRYPT_ENABLE to MK_TME_ENABLE
- Add check production to MSR section
- Nicely handle the case where mpa_registration.log is not available

Fixes #192

### DIFF
--- a/system-report.sh
+++ b/system-report.sh
@@ -106,10 +106,6 @@ print_section "Model specific registers (MSRs)" "${result}"
 result=$(grep -m1 "model name" /proc/cpuinfo | cut -f2 -d":")
 print_section "CPU details" "${result}"
 
-result=$(find . -name check-production.sh -exec \
-    sh -c 'cd "$(dirname "$0")" && sudo ./check-production.sh' {} \;)
-print_section "Production system check" "${result}"
-
 set_pkg_result_string "qemu-system-x86"
 print_section "QEMU package details" "${result}"
 
--- a/system-report.sh
+++ b/system-report.sh
@@ -60,8 +60,8 @@ install_msr() {
 
 set_msr_result_string() {
     install_msr
-    MK_TME_ENABLE=$(sudo rdmsr 0x982 -f 1:1)
-    result="MK_TME_ENABLE bit: ${MK_TME_ENABLE} (expected value: 1)"
+    MK_TME_ENABLED=$(sudo rdmsr 0x982 -f 1:1)
+    result="MK_TME_ENABLED bit: ${MK_TME_ENABLE} (expected value: 1)"
     SEAM_RR=$(sudo rdmsr 0x1401 -f 11:11)
     result="$result\nSEAM_RR bit: $SEAM_RR (expected value: 1)"
     NUM_TDX_PRIV_KEYS=$(sudo rdmsr 0x87 -f 63:32)
--- a/system-report.sh
+++ b/system-report.sh
@@ -50,15 +50,30 @@ set_pkg_result_string() {
     result="$result\n$(apt info ${package} 2>/dev/null | grep -E 'Package|Version|APT-Sources')"
 }
 
+# load msr kernel module and install msr-tools if needed
+install_msr() {
+    if ! which rdmsr &> /dev/null; then
+        sudo apt install -y msr-tools &> /dev/null
+    fi
+    sudo modprobe msr &> /dev/null
+}
+
 set_msr_result_string() {
-    HW_ENCRYPT_ENABLE=$(sudo rdmsr 0x982 -f 1:1)
-    result="HW_ENCRYPT_ENABLE bit: ${HW_ENCRYPT_ENABLE} (expected value: 1)"
+    install_msr
+    MK_TME_ENABLE=$(sudo rdmsr 0x982 -f 1:1)
+    result="MK_TME_ENABLE bit: ${MK_TME_ENABLE} (expected value: 1)"
     SEAM_RR=$(sudo rdmsr 0x1401 -f 11:11)
     result="$result\nSEAM_RR bit: $SEAM_RR (expected value: 1)"
     NUM_TDX_PRIV_KEYS=$(sudo rdmsr 0x87 -f 63:32)
-    result="$result\nNUM_TDX_PRIV_KEYS: $NUM_TDX_PRIV_KEYS (expected value: >32)"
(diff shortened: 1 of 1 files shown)


system-report.sh
```diff
@@ -50,15 +50,30 @@ set_pkg_result_string() {
...
+    MK_TME_ENABLE=$(sudo rdmsr 0x982 -f 1:1)
+    result="MK_TME_ENABLE bit: ${MK_TME_ENABLE} (expected value: 1)"
     SEAM_RR=$(sudo rdmsr 0x1401 -f 11:11)
     result="$result\nSEAM_RR bit: $SEAM_RR (expected value: 1)"
     NUM_TDX_PRIV_KEYS=$(sudo rdmsr 0x87 -f 63:32)
-    result="$result\nNUM_TDX_PRIV_KEYS: $NUM_TDX_PRIV_KEYS (expected value: >32)"
+    result="$result\nNUM_TDX_PRIV_KEYS: $NUM_TDX_PRIV_KEYS"
     SGX_AND_MCHECK_STATUS=$(sudo rdmsr 0xa0)
     result="$result\nSGX_AND_MCHECK_STATUS: $SGX_AND_MCHECK_STATUS (expected value: 0)"
+    PROD=$(rdmsr 0xce -f 27:27)
```
frenchwr, line 71: I think this is redundant since we're already doing this by running the `check-production.sh` script [here](https://github.com/canonical/tdx/blob/558594ea444286bf198a26df1acdbe9d2d9895e2/system-report.sh#L109-L111). So either this should (a) be moved to the section below and we should not be running `check-production.sh` in that section, or (b) we should remove the dedicated section for the production checking. I'm fine either way. 

I suppose another option would be to just migrate the whole `check-production.sh` script contents into `system-check.sh` and remove `check-production.sh` entirely. Although maybe it's nice to only check for production with all the other details about the system.

frenchwr (changes requested):
Looks good - one comment about the production check.

system-report.sh
hector-cao: @frenchwr Ah; i did not get that we already call `check-production.sh` script when i first read the issue #192

hector-cao: I asked the issue reporter for more clarification but in the meantime, i will go for option b) for now to make the script as standalone as possible

BFuhry: @frenchwr I was not aware that there is a discussion here about this topic. Please see issue #192 for my comments about the `check-production.sh` script.

system-report.sh
```diff
@@ -50,15 +50,30 @@ set_pkg_result_string() {
...
+    fi
+    sudo modprobe msr &> /dev/null
+}
+
 set_msr_result_string() {
-    HW_ENCRYPT_ENABLE=$(sudo rdmsr 0x982 -f 1:1)
-    result="HW_ENCRYPT_ENABLE bit: ${HW_ENCRYPT_ENABLE} (expected value: 1)"
+    install_msr
+    MK_TME_ENABLE=$(sudo rdmsr 0x982 -f 1:1)
+    result="MK_TME_ENABLE bit: ${MK_TME_ENABLE} (expected value: 1)"
```
BFuhry, line 64:
```suggestion
    MK_TME_ENABLED=$(sudo rdmsr 0x982 -f 1:1)
    result="MK_TME_ENABLED bit: ${MK_TME_ENABLE} (expected value: 1)"
```

BFuhry (commented):
Minor suggestion about name of bit. Besides, I would move the CPU platform recognition from `check-production.sh` to here. From the script name, it does not belong there, but it might be helpful here.

system-report.sh
frenchwr: @BFuhry Thanks, I was following along. Sometimes I lurk. :)

frenchwr (approved):
LGTM!

PR #197, opened 2024-08-22
create-td-image : use qemu virt type for virt-install if kvm is missing

hector-cao:
if we run the create-td-image.sh in a system that lacks KVM support, the image creation fails (`virt-install` tool fails)
virt-install has support to use qemu instead of kvm; so use it as a fallback since qemu offers a lower performance than kvm (on a system, we spend 2m to create an image with KVM and 7m to create an image with QEMU)


bktan8 (approved):
LGTM

bktan8 (approved):

PR #217, opened 2024-09-12
kvm_intel module comes with tdx enabled by default

hector-cao:
we do not need to setup grub to enable tdx argument for kvm_intel anymore

### DIFF
--- a/setup-tdx-host.sh
+++ b/setup-tdx-host.sh
@@ -110,7 +110,10 @@ grub_set_kernel
 # just force the installation of modules-extra to make sure we have it
 apt install --yes --allow-downgrades linux-modules-extra-${KERNEL_RELEASE}
 
-grub_cmdline_kvm || true
+# in recent -intel kernel, tdx is enabled by default in kvm_intel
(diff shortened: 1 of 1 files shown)


setup-tdx-host.sh
```diff
@@ -110,7 +110,10 @@ grub_set_kernel
 # just force the installation of modules-extra to make sure we have it
 apt install --yes --allow-downgrades linux-modules-extra-${KERNEL_RELEASE}
 
-grub_cmdline_kvm || true
+# in recent -intel kernel, tdx is enabled by default in kvm_intel
+# so this is not necessary anymore
+#grub_cmdline_kvm || true
```
ThibF, line 115: I'm in favor to just remove dead code, but if you feel strongly to keep it, that's okay for me.

ThibF (approved):

setup-tdx-host.sh
hector-cao: Yeah, agree, however i kept it to help users to understand this transition without having to read git history, will remove it later

PR #219, opened 2024-09-12
PEK 1233 Add tests as aligned upon with Intel

spmcmillan:
https://warthogs.atlassian.net/browse/PEK-1233

Created the following tests:
- tdx_4vcpus_2sockets_5times
- tdx_4vcpus_1socket_10times
- TDX_debugoff_earlyprink
- tdx_max_guests
- td_max_vcpus
- tdx_QGS_socket

Reorganized and removed a duplicate test.


tests/tests/test_resource_limits.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def test_max_guests():
+    """
+    Test max guests (No Intel Case ID)
+    """
+
+    # get max number of TD VMs we can create (max - current)
+    max_td_vms = util.get_max_td_vms() - util.get_current_td_vms()
+    qm = [None] * max_td_vms
```
hector-cao, line 68: add `assert` here to make sure we have max_td_vms  > 0 ?

tests/tests/test_resource_limits.py
```diff
@@ -0,0 +1,89 @@
...
+import Qemu
+import util
+
+def test_huge_resource_vm():
+    """
+    Test huge resources  (Intel Case ID 007)
+    """
+
+    # choose half available memory per Intel case spec
+    huge_mem_gb = int(util.get_memory_free_gb() / 2)
```
hector-cao, line 29: here we are creating VM with half of free memory
```
*               total        used        free      shared  buff/cache   available
Mem:            13Gi       6.6Gi       1.5Gi       211Mi       5.9Gi       6.9Gi
Swap:          4.0Gi       1.0Mi       4.0Gi
```
for this example, the free memory is `1.5Gb`.
We are following the instructions at : https://github.com/intel/tdx/wiki/Tests
however, i believe we could do better, at least we can use the `available` or even `total` value

hector-cao (changes requested):
Great work ! one general remark is i would like we use `ssh` access to check VM being functional.

tests/tests/test_boot_basic.py
```diff
@@ -37,3 +37,18 @@ def test_guest_boot():
...
 
     qm.stop()
+
+def test_early_printk():
+    """
+    Test Early Printk with Debug Off (Intel Case ID 018)
+    """
+
+    qm = Qemu.QemuMachine()
+    qm.qcmd.plugins['boot'].kernel = "/boot/vmlinuz"
```
hector-cao, line 47: Here we are using the host kernel that can be different from the guest kernel (that is inside the image). I dont think we should follow "stricto sensu" the instructions from Intel. The idea here is to only add `earlyprintk=ttyS0,115200` to kernel commandline

hector-cao (changes requested):

tests/tests/test_boot_basic.py
spmcmillan: You can't add earlyprintk without kernel.  That is what I tried first.

tests/tests/test_multiple_boot.py
```diff
@@ -20,15 +20,37 @@
...
     qm = Qemu.QemuMachine()
+    qm.qcmd.plugins['cpu'].nb_cores = 4
+    qm.qcmd.plugins['cpu'].nb_sockets = 2
+
+    Qemu.QemuMonitor.CONNECT_RETRIES = 10
 
     for i in range(0,5):
         qm.run()
-        m = Qemu.QemuSSH(qm)
+        mon = Qemu.QemuMonitor(qm)
```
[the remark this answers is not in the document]
spmcmillan: If I don't need SSH, I used monitor because its faster.  However, I do think SSH makes sense to wait for everything to cleanly boot up and will change it.

tests/tests/test_quote_configfs_tsm.py
```diff
@@ -36,3 +35,25 @@ def test_quote_check_configfs_tsm():
...
+
+    qm = Qemu.QemuMachine()
+    qm.run()
+
+    mon = Qemu.QemuMonitor(qm)
+    mon.wait_for_state('running')
+
+    # do basic tsm_config test on guest
+    ssh = Qemu.QemuSSH(qm)
+    ssh.rsync_file(f'{script_path}/../lib', '/tmp/tdxtest/')
```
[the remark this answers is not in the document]
spmcmillan: Sure, that wasn't available previously.  I'll merge main

tests/tests/test_multiple_boot.py
```diff
@@ -20,15 +20,37 @@
...
+def test_4vcpus_1socket_10times():
     """
-    Boot TD several times
+    Test 4vcpus 1socket 10 times (Intel Case ID 009)
     """
+
+    qm = Qemu.QemuMachine()
+    qm.qcmd.plugins['cpu'].nb_cores = 4
+
+    Qemu.QemuMonitor.CONNECT_RETRIES = 10
```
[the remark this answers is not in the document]
spmcmillan: I believe the wait seemed excessive for something that should be done quickly.  But, I'l change to SSH and this won't matter

tests/tests/test_multiple_boot.py
```diff
@@ -20,15 +20,37 @@
 import os
 import Qemu
 
-script_path=os.path.dirname(os.path.realpath(__file__))
```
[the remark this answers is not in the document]
spmcmillan: I was trying to name more closely to Intel, but probably failed that as well.  It wasn't really a "reboot".  But, I'm happy to try to change the name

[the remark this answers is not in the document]
hector-cao: you are right that the old name is not good, i propose `test_boot_multiple` or something similar

tests/tests/test_multiple_boot.py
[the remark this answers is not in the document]
hector-cao: IIRC, initially I set this RETRIES value to 10 and on some systems that has old firmware, i had failures something, maybe i was doing something wrong to have this failure but i ended up somehow with the value 60

tests/tests/test_boot_basic.py
hector-cao: ah, I see that it is not a trivial problem, let do it in a following PR and have some time to think about it, it will allow us to wrap up this PR quickly

tests/tests/test_resource_limits.py
spmcmillan: I think if we us total or available we could end up asking for more than is available to use.  It seems we should use either half of free or some other percentage.  We can always add another one that adds 75% of free or 90% of free or even 99% of free....

hector-cao: for `total`, i have to check how qemu asks for memory (let me do a quick check), but for `available`, i think that should be fine (https://askubuntu.com/questions/867068/what-is-available-memory-while-using-free-command).

hector-cao (changes requested):

tests/tests/test_resource_limits.py
hector-cao: https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/7/html/virtualization_deployment_and_administration_guide/chap-overcommitting_with_kvm

hector-cao: Just did the test, i can start a VM with 125G RAM when my host has only less than 70G available

image

spmcmillan: I added another test to hit this available limit.  Keeping the "intel" test the same.

tests/tests/test_stress_resources.py
```diff
@@ -0,0 +1,104 @@
...
+    ssh = Qemu.QemuSSH(qm)
+
+    qm.stop()
+
+def test_memory_limit_resource_vm():
+    """
+    Test memory limit resource  (No Intel Case)
+    """
+
+    # choose half available memory per Intel case spec
```
hector-cao, line 47:
```suggestion
    # run VM with all available memory
```

tests/tests/test_stress_resources.py
```diff
@@ -0,0 +1,104 @@
...
+
+import Qemu
+import util
+
+def test_huge_resource_vm():
+    """
+    Test huge resources  (Intel Case ID 007)
+    """
+
+    # choose half available memory per Intel case spec
```
hector-cao, line 29:
```suggestion
    # run VM with half of the free memory
```

hector-cao (approved):
LGTM, minor suggestions

PR #221, opened 2024-09-13
Improve the management of workdir cleanup for debugging purposes

hector-cao:
- In case of test failure, we would like that the workdir of qemu is not cleaned up so that we can do some debugging.
- Improve the management of environment variable TDXTEST_DEBUG that forbids the clean up of these workdirs at any case

### DIFF
--- a/tests/tests/conftest.py
+++ b/tests/tests/conftest.py
@@ -1,0 +1,28 @@
+import os
+import pytest
+
+import Qemu
+
+@pytest.fixture(autouse=True)
+def run_before_and_after_tests(tmpdir):
+    """
+    Fixture to execute before and after a test is run
+    Even in case of test failure
+    """
+    # Setup: fill with any logic you want
+    # enable the debug flag if TDXTEST_DEBUG is set
+    debug = os.environ.get('TDXTEST_DEBUG', False)
+    if debug:
+        Qemu.QemuMachine.set_debug(debug)
+
+    yield # this is where the testing happens
+
+    # Teardown : fill with any logic you want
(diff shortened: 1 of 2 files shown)


tests/tests/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+
+    # Teardown : fill with any logic you want
+
+def pytest_exception_interact(node, call, report):
+    """
+    Called at test failure
+    """
+    if report.failed:
+        # enable debug flag to avoid cleanup to happen
+        Qemu.QemuMachine.set_debug(True)
```
spmcmillan, line 28: why do we want to avoid cleanup when a test fails?  Don't we want the next text to start cleanly?  I understand if we running individually and we can set "debug", but not generally

spmcmillan (commented):
Hi Hector, overall looks good.  Added a comment.  One other comment, another option when terminating Qemu is to add the kill() signal after trying to terminate properly. (in stop())

tests/tests/conftest.py
hector-cao: each qemu has its own working folder (/tmp/tdxtest-xxxxx/) that is not shared with other tests qemu instances

hector-cao:
> Hi Hector, overall looks good. Added a comment. One other comment, another option when terminating Qemu is to add the kill() signal after trying to terminate properly. (in stop())

I decided to split this MR into 2, the current one is now only about workdir clean up and debug flag

spmcmillan (approved):
LGTM

PR #222, opened 2024-09-16
Use GiB as unit for memory conf in libvirt configuration

hector-cao:
It is easier to work with Gb unit, this change is to ease the users when they want to quickly configure the VM memory to an other value for testing


frenchwr:
Seems fine but could you provide some context and motivation for the change?

hector-cao:
> Seems fine but could you provide some context and motivation for the change?

I just wrote something in the MR description, tell me if that makes sense

frenchwr (approved):

PR #229, opened 2024-09-20
Pek614 adding attestation type tests

spmcmillan:
https://warthogs.atlassian.net/browse/PEK-614

Adding trustauthority-cli quote test.

### DIFF
--- a/tests/tests/test_guest_measurement.py
+++ b/tests/tests/test_guest_measurement.py
@@ -1,8 +1,6 @@
 #!/usr/bin/env python3
 #
 # Copyright 2024 Canonical Ltd.
-# Authors:
-# - Hector Cao <hector.cao@canonical.com>
 #
 # This program is free software: you can redistribute it and/or modify it
 # under the terms of the GNU General Public License version 3, as published
@@ -19,6 +17,8 @@
 
 import os
 import time
+import json
+import subprocess
 
 import Qemu
 import util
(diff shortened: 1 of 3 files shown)


tests/tests/test_guest_measurement.py
```diff
@@ -28,13 +28,58 @@ def test_guest_measurement_check_rtmr(qm):
...
+    Qemu.QemuMachineType.Qemu_Machine_Params[Qemu.QemuEfiMachine.OVMF_Q35_TDX][1] = object
+
+    quote_str = ""
+    with Qemu.QemuMachine() as qm:
+        qm.qcmd.add_vsock(3)
+        qm.run()
+
+        ssh = Qemu.QemuSSH(qm)
+
+        deploy_and_setup(ssh)
```
hector-cao, line 81: Very minor but I think for this test, we do not need deploy_and_setup() action

tests/tests/test_guest_measurement.py
```diff
@@ -28,13 +28,58 @@ def test_guest_measurement_check_rtmr(qm):
...
-    deploy_and_setup(m)
+        deploy_and_setup(m)
 
-    m.check_exec('tdrtmrcheck')
+        m.check_exec('tdrtmrcheck')
 
-    qm.stop()
+        qm.stop()
 
+def test_guest_measurement_trust_authority_success():
```
hector-cao, line 42: i m wondering if we have to put these tests in a separate file, this file is initially reserved for tests of measurement values. i would suggest the new file name `test_guest_attestation_ita` or `test_guest_attestation_quote`

hector-cao (approved):
LGTM, some nit comments

PR #230, opened 2024-09-20
test_guest_cpu_off: fixes failure

hector-cao:
- if we pin the process to a offline cpu core, the test will fail we need to bring the core online fist
- add a cpu manager as context manager to easily restore the core state at the end of the test
- move some utility code to util.py
- make cpu random select more robust since on some systems core0 might not support the on/off feature (IRQ is wired on it)

### DIFF
--- a/tests/tests/test_guest_cpu_off.py
+++ b/tests/tests/test_guest_cpu_off.py
@@ -21,15 +21,13 @@ import os
 import subprocess
 import time
 import re
-import multiprocessing
-import random
 
 import Qemu
 import util
 
 script_path=os.path.dirname(os.path.realpath(__file__))
 
-def test_guest_cpu_off(qm):
+def test_guest_cpu_off(qm, cpu_core):
     """
     tdx_VMP_cpu_onoff test case (See https://github.com/intel/tdx/wiki/Tests)
     """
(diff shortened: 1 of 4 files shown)


tests/tests/test_guest_cpu_off.py
```diff
@@ -67,39 +55,16 @@ def test_guest_cpu_pinned_off():
...
+                # bring the cpu online if necessary
+                cpu_manager.set_state(1)
+                util.pin_process_on_cpu(qm.pid, cpu)
 
-            m = Qemu.QemuSSH(qm)
-            m.check_exec('sudo init 0 &')
+                # bring the cpu core offline
+                cpu_manager.set_state(0)
+                m = Qemu.QemuSSH(qm)
+                m.check_exec('sudo init 0 &')
```
spmcmillan, line 68: Small nit, I think its worth explicitly setting state back to 1 even if the context manager sets it .  When someone looks at this code they may not understand what is happening when the object gets destroyed at first and personally, I prefer explicit and then using implicit (context manager) as a backup.

hector-cao: @spmcmillan  I see, however, we do not want to set the core state to 1, but to the previous state before the test (for example, the core can be offline just before the test), the context manager does it for us right now.

spmcmillan: I think you are saying its possible the cpu came in turned off already.  Which I suppose is true

PR #231, opened 2024-09-20
test_guest_early_printk : fix test

hector-cao:
for unknown reason, running the qemu with Popen() without shell=True will run the qemu but ssh access is impossible
when we run the same qemu commandline from the terminal, the VM starts fine and ssh access is available

### DIFF
--- a/tests/lib/Qemu.py
+++ b/tests/lib/Qemu.py
@@ -534,32 +534,36 @@ class QemuMachine:
                               shell=True,
                               stdout=subprocess.DEVNULL)
 
-    def run(self):
+    def run(self, shell=False):
         """
         Run qemu
         """
         cmd = self.qcmd.get_command()
-        print(' '.join(cmd))
-        self.proc = subprocess.Popen(cmd,
-                                    stdout=subprocess.PIPE,
-                                    stderr=subprocess.PIPE)
+        cmd_str = ' '.join(cmd)
+        print(cmd_str)
+        # if shell=True is enabled, we replace the shell with qemu process
+        # this will allow to properly terminate the qemu process and shell
+        # with self.proc.terminate()
+        # otherwise, the shell process still live as zombie and block
+        # the collection of logs with communicate()
+        if shell:
+            cmd = f'exec {cmd_str}'
+        self.proc = subprocess.Popen(cmd, shell=shell,
+                                     stdout=subprocess.PIPE,
+                                     stderr=subprocess.PIPE)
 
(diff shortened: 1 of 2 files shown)


spmcmillan (approved):
LGTM.  This issue isn't ssh, but that its not booting.  Maybe an environment variable but very weird.  TDX not letting it boot without shell?  Here is some debug I did FYI...

For some reason, the system does not actually bootup without shell=True.  If you change it back to shell=False add the line for initrd:

qm.qcmd.plugins['boot'].initrd = "/boot/initrd.img"

And then, add "-serial /dev/pts/1" (for me "who" reported i was on pts/1

I did that in Qemu.py just temporarily so I could see whats going on:
    def get_command(self):
        _args = ['qemu-system-x86_64']
        for p in self.plugins.values():
            _args.extend(p.args())
        return _args + self.command + ['-serial', '/dev/pts/1']

You'll see it not booting up and see only the following two lines:
EFI stub: Loaded initrd from LINUX_EFI_INITRD_MEDIA_GUID device path
EFI stub: Measured initrd data into PCR 9

If you change "shell=True" back you'll see:
EFI stub: Loaded initrd from LINUX_EFI_INITRD_MEDIA_GUID device path
EFI stub: Measured initrd data into PCR 9
[    0.000000] tdx: Guest detected
[    0.000000] Linux version 6.8.0-45-generic (buildd@lcy02-amd64-115) (x86_64-linux-gnu-gcc-13 (Ubuntu 13.2.0-23ubuntu4) 13.2.0, GNU ld (GNU Binutils for Ubuntu) 2.42) #45-Ubuntu SMP PREEMPT_DYNAMIC Fri Aug 30 12:02:04 UTC 2024 (Ubuntu 6.8.0-45.45-generic 6.8.12)
[    0.000000] Command line: root=/dev/vda1 earlyprintk=ttyS0,115200 initrd=initrd
[    0.000000] KERNEL supported cpus:
[    0.000000]   Intel GenuineIntel
[    0.000000]   AMD AuthenticAMD

spmcmillan:
You don't need this if we just use https://github.com/canonical/tdx/commit/8bba988aa59c3df7545b131e0b81083eb0a2b53b

hector-cao:
> You don't need this if we just use [8bba988](https://github.com/canonical/tdx/commit/8bba988aa59c3df7545b131e0b81083eb0a2b53b)

Great ! i think your proposal is good, i close this PR since it is no longer needed

PR #232, opened 2024-09-23
test_guest_early_printk : test with the kernel in the guest VM

hector-cao:
right now, we are testing the host kernel
actually, we want to do this test for the guest VM kernel
I propose to fix that by redesigning the test in 2 steps:
  - append the wanted kernel commandline in the guest VM
  - restart the guest VM

### DIFF
--- a/tests/lib/Qemu.py
+++ b/tests/lib/Qemu.py
@@ -578,6 +578,19 @@ class QemuMachine:
             except Exception as e:
                 print(f'Exception {e}')
 
+    def reboot(self):
+        """
+        Reboot the QEMU machine
+        Since VM might not support reboot just poweroff and start gain
+        """
+        m = QemuSSH(self)
+        # sync data and poweroff
+        m.check_exec('sync && systemctl poweroff')
+        # check that VM quits properly
+        self.communicate()
+        # run the VM again
+        self.run()
+
     def __del__(self):
         """
         Make sure we stop the qemu process if it is still running
--- a/tests/tests/test_boot_basic.py
+++ b/tests/tests/test_boot_basic.py
@@ -41,10 +41,19 @@ def test_guest_early_printk(qm):
     Test Early Printk with Debug Off (Intel Case ID 018)
     """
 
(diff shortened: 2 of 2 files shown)


spmcmillan (commented):
This makes sense, however, it changes the image permanently and might affect other tests.  Should we go through and modify it back?  Though if this test fails, we cannot modify it back.  Which is not great.

fanchen2:
> This makes sense, however, it changes the image permanently and might affect other tests. Should we go through and modify it back? Though if this test fails, we cannot modify it back. Which is not great.

We often add "earlyprintk" in kernel cmdline to show more logs when booting, this is an old common function of kernel, which can be token as a common setting, just FYI.

fanchen2:
> > > This makes sense, however, it changes the image permanently and might affect other tests. Should we go through and modify it back? Though if this test fails, we cannot modify it back. Which is not great.
> > 
> > 
> > We often add "earlyprintk" in kernel cmdline to show more logs when booting, this is an old common function of kernel, which can be token as a common setting, just FYI.
> 
> Thanks @fanchen2 , does the existence of this test imply that this has failed for TD in the past?

The original case is to check if TDVM can boot up with debug=off and earlyprintk, the previous failure is caused by "debug" function, which is not integrated into this software stack. I understand you, we expect earlyprintk works all the time, but on one can ensure it.

spmcmillan:
> > This makes sense, however, it changes the image permanently and might affect other tests. Should we go through and modify it back? Though if this test fails, we cannot modify it back. Which is not great.
> > A really ugly way is to use qemu-nbd, mount the image, pull out the kernel, unmount the image and use the pulled out kernel...
> 
> we designed the tests so that each QEMU machine has its own image layer for its modifications (through overlay mechanism : https://libvirt.org/kbase/backing_chains.html ). So, on top of the original image, each QEMU machine will create an overlay image that sits on top of the original one and all the modifications to the machine will only live in the overlay layer and not go into the original image

Good deal.  LGTM

spmcmillan (approved):
LGTM

PR #234, opened 2024-09-25
system-report : fix too long output for GH issue

hector-cao:
GH issue accepts comments with length less than 65k characters with dmesg output, if the machine is up and running for a quite long period it can generate very long  output

To mitigate this issue, we limits the dmesg output to the first 10 lines and 20 last lines only. The first 10 lines allow us to see the tdx module load and the 20 last lines might show us the log that happens during the issue


spmcmillan (approved):
LGTM

PR #235, opened 2024-09-25
Adding memory confidentiality tests

spmcmillan:
JIRA: https://warthogs.atlassian.net/browse/PEK-1285

Adding tests to verify confidentiality with TDX and verify lack of confidentiality without TDX.

### DIFF
--- a/tests/tests/test_guest_memory.py
+++ b/tests/tests/test_guest_memory.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python3
+#
+# Copyright 2024 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 3, as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+# SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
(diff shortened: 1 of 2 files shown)


tests/tests/test_guest_memory.py
```diff
@@ -0,0 +1,81 @@
...
+import subprocess
+
+import Qemu
+from common import *
+
+MEMORY_FILE='memory.bin'
+CHECK_EXEC='export INCLEAR=6150079244'
+
+def check_for_string(filename, needle):
+    cs = subprocess.run(['sudo', 'grep', needle, filename], check=True, capture_output=True)
```
hector-cao, line 30: with `check=True`, the call will throw an exception if the string is not present in the file, that will make the test `test_guest_memory_confidentiality_tdx` fails and we do not want that, do we ?

hector-cao (changes requested):

tests/tests/test_guest_memory.py
spmcmillan: Yes, thanks for catching that.

spmcmillan: fixed in https://github.com/canonical/tdx/pull/235/commits/ea47d4ddcd72bba77caaf26e041bcc8cf3132d0e

hector-cao (commented):
LGTM, do you plan to add these tests into checkbox in this PR or in an other upcoming PR ?

spmcmillan:
> LGTM, do you plan to add these tests into checkbox in this PR or in an other upcoming PR ?

Just added it and tested.

hector-cao (approved):
LGTM, awesome work

PR #236, opened 2024-09-25
Add GNR to list of supported CPUs in Readme

bktan8:
Remove SRF from check_production.sh since it's not launched yet.


hector-cao (approved):
LGTM

PR #237, opened 2024-09-29
Kexec/Kdump Tests

spmcmillan:
https://warthogs.atlassian.net/browse/PEK-852

Adding kexec/kdump tests for the host.  This requires the host to be rebooted effectively multiple times.  So, this was written as a shell script that can be used, eventually, in testflinger, or in a runner.

### DIFF
--- a/tests/tests/test_host_kdump.sh
+++ b/tests/tests/test_host_kdump.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/bash
+#
+# Copyright 2024 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 3, as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+# SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+# sudo apt install linux-crashdump
+# reboot
+# kdump-config show
+# cat /proc/cmdline --> verify crashkernel value in place 
+# validate cat /proc/sys/kernel/sysrq > 0
+# trigger crash --> sudo "echo c > /proc/sysrq-trigger"
+# find crashdump in /var/crash
+
+if [ ! -v DEVICE_IP ]; then
+    DEVICE_IP="192.168.102.125"
+fi
+
+echo "Installing linux-crashdump and rebooting"
+ssh ubuntu@$DEVICE_IP sudo apt install linux-crashdump
+if [ $? != 0 ]; then
+    echo "Can't install linux-crashdump"
+    exit -1
+fi
+
+ssh ubuntu@$DEVICE_IP sudo systemctl reboot
+if [ $? != 0 ]; then
+    echo "Can't reboot"
+    exit -2
+fi
+
+sleep 30
+echo "Waiting for system to come back up"
+
+cnt=0
+until ssh ubuntu@$DEVICE_IP ls &> /dev/null; do sleep 1; cnt=$(expr $cnt + 1); if [ $cnt -gt 120 ]; then break; fi; done
+if [ $cnt -gt 60 ]; then
+    echo "Timed out waiting for $DEVICE_IP to come back up ($cnt)"
+    exit -3
+fi
+if [ $cnt == 0 ]; then
+    echo "$DEVICE_IP came back too quickly"
+    exit -4
+fi
+
+echo "System came back up"
+
+KDUMP_CONFIG="$(ssh ubuntu@$DEVICE_IP sudo kdump-config show)"
+if [ $? != 0 ]; then
+    echo "Failed getting kdump config"
+    exit -5
+fi
+
+CMDLINE="$(ssh ubuntu@$DEVICE_IP cat /proc/cmdline)"
+if [ $? != 0 ]; then
+    echo "Failed getting cmd line"
+    exit -6
+fi
+
+CRASH_DIR="$(ssh ubuntu@$DEVICE_IP ls /var/crash)"
+if [ $? != 0 ]; then
+    echo "Failed getting crash directory"
+    exit -7
+fi
+
+
+echo "Crashing system"
+ssh ubuntu@$DEVICE_IP "echo c | sudo tee /proc/sysrq-trigger" &
+
+sleep 10
+echo "Waiting for system to come back up"
+
+cnt=0
+until ssh ubuntu@$DEVICE_IP ls &> /dev/null; do sleep 1; cnt=$(expr $cnt + 1); if [ $cnt -gt 120 ]; then break; fi; done
+if [ $cnt -gt 60 ]; then
+    echo "Timed out waiting for $DEVICE_IP to come back up ($cnt)"
+    exit -9
+fi
+if [ $cnt == 0 ]; then
+    echo "$DEVICE_IP came back too quickly"
+    exit -10
+fi
+
+echo "System came back up"
+CRASH_DIR2="$(ssh ubuntu@$DEVICE_IP ls /var/crash)"
+if [ $? != 0 ]; then
+    echo "Failed getting crash directory"
+    exit -11
+fi
+
+echo $CRASH_DIR
+echo "----------------------------------------------------"
+echo "----------------------------------------------------"
+echo "----------------------------------------------------"
+echo $CRASH_DIR2
--- a/tests/tests/test_host_kdump.sh
+++ b/tests/tests/test_host_kdump.sh
@@ -17,7 +17,7 @@
 
 if [ ! -v DEVICE_IP ]; then
     echo "Must define DEVICE_IP"
-    exit -13
+    exit -1
 fi
 
 echo "Installing linux-crashdump and rebooting"
@@ -30,7 +30,7 @@ fi
 ssh ubuntu@$DEVICE_IP sudo systemctl reboot
 if [ $? != 0 ]; then
     echo "Can't reboot"
-    exit -2
+    exit -1
 fi
 
 sleep 15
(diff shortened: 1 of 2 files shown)


frenchwr:
@spmcmillan Just making sure I understand before looking at the code - are these the first tests that need to be run remotely over `ssh` to the TDX host? If so do they need some sort of special treatment or category that is separate from existing `test_host*` tests (since the existing tests need to be run directly from the host while these new ones need to be run from a machine with ssh access to the host)? I guess eventually maybe all the tests could be run this way by invoking commands remotely over `ssh`?

spmcmillan:
> @spmcmillan Just making sure I understand before looking at the code - are these the first tests that need to be run remotely over `ssh` to the TDX host? If so do they need some sort of special treatment or category that is separate from existing `test_host*` tests (since the existing tests need to be run directly from the host while these new ones need to be run from a machine with ssh access to the host)? I guess eventually maybe all the tests could be run this way by invoking commands remotely over `ssh`?

Yes, these are probably the only tests that will need to be run remotely as they require the system to reboot or crash whenever they are invoked.  I'm not sure the ideal place to put these.  They won't be picked up by the pytests so they are already isolated and won't be included with the checkbox tests either.  They are very special case and I can see them running on testflinger on the agent.  I thought this was an ok to put them because they have the ".sh" extension.  But, if you have other ideas please suggest them!

frenchwr:
>  They won't be picked up by the pytests so they are already isolated and won't be included with the checkbox tests either.

Ah ok - I was thinking they would but I guess with the .sh extension tox/pytest will not pick them up. :+1:

tests/tests/test_host_kdump.sh
```diff
@@ -0,0 +1,103 @@
...
+if [ $? != 0 ]; then
+    echo "Can't reboot"
+    exit -2
+fi
+
+sleep 30
+echo "Waiting for system to come back up"
+
+cnt=0
+until ssh ubuntu@$DEVICE_IP ls &> /dev/null; do sleep 1; cnt=$(expr $cnt + 1); if [ $cnt -gt 120 ]; then break; fi; done
```
frenchwr, line 39: I have a few questions about this block:

1. Why increment to 120 but then check against 60 in the first if statement below? Shouldn't the two numbers match? Or can we just exit from the if statement inside the `until` block and then remove the `if [ $cnt -gt 60 ]` check below?
2. If I'm reading this correctly, we expect after the reboot that ssh-server will be up on the TDX host within 30 - 90 seconds (I guess that assumes that ssh returns immediately and does not hang). Is that actually a specific performance requirement some place?

tests/tests/test_host_kdump.sh
```diff
@@ -0,0 +1,103 @@
...
+if [ $cnt -gt 60 ]; then
+    echo "Timed out waiting for $DEVICE_IP to come back up ($cnt)"
+    exit -3
+fi
+if [ $cnt == 0 ]; then
```
frenchwr, line 44: Just curious why this is needed? Have you seen the system boot up quickly and that being indicative of a problem?

tests/tests/test_host_kdump.sh
```diff
@@ -0,0 +1,103 @@
...
+fi
+
+echo "Crashing system"
+ssh ubuntu@$DEVICE_IP "echo c | sudo tee /proc/sysrq-trigger" > /dev/null &
+
+sleep 10
+echo "Waiting for system to come back up"
+
+cnt=0
+until ssh ubuntu@$DEVICE_IP ls &> /dev/null; do sleep 1; cnt=$(expr $cnt + 1); if [ $cnt -gt 120 ]; then break; fi; done
```
frenchwr, line 77: Same comments as above. Also since we're doing this again should we wrap it in a function?

tests/tests/test_host_kdump.sh
```diff
@@ -0,0 +1,103 @@
...
+
+if [ ! -v DEVICE_IP ]; then
+    DEVICE_IP="192.168.102.125"
+fi
+
+echo "Installing linux-crashdump and rebooting"
+ssh ubuntu@$DEVICE_IP sudo apt install linux-crashdump
+if [ $? != 0 ]; then
+    echo "Can't install linux-crashdump"
+    exit -1
```
frenchwr, line 26: I wonder how useful it is to use custom shell return codes for each failure type here? Could we instead just use the return code from the ssh command?

spmcmillan: I'm not sure its ever useful (error codes in bash scripts), but I like to give someone the option of figuring out where it failed given a return code.  I think often people will either throw and exception in these cases with a message only (in python).  I don't think the return code from ssh is useful to figure out where it failed.  I think these things are dilemmas.  I prefer this way, but I certainly understand why an exception would be great.  In this case we'll definitely get any error codes in stdout.

tests/tests/test_host_kdump.sh
spmcmillan: The whole point of these tests is to ensure reboots, kexec, and crash (with kdump/kexec) work properly.   So, we want to make sure it actually rebooted/crashed, ....  If it doesn't actually reboot/crash/..., then its not really testing this properly.  So, its a failure if it doesn't actually reboot.

tests/tests/test_host_kdump.sh
spmcmillan: No performance requirement, just felt right.  The 120/60 thing was a mistake on my part.  I thought 120 was better.  In general, I don't want to fail if it just needs a little longer to boot.  Good find.  I'll fix the 120/60 thing.

tests/tests/test_host_kdump.sh
```diff
@@ -0,0 +1,103 @@
...
+# ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+# SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+if [ ! -v DEVICE_IP ]; then
+    DEVICE_IP="192.168.102.125"
```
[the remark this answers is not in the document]
spmcmillan: Thats probably a good idea so we don't accidentally crash the wrong device :-)

tests/tests/test_host_kdump.sh
spmcmillan: https://github.com/canonical/tdx/pull/237/commits/49834af5bd0588025ff659bb50199732297263c7

tests/tests/test_host_kdump.sh
spmcmillan: https://github.com/canonical/tdx/pull/237/commits/49834af5bd0588025ff659bb50199732297263c7

tests/tests/test_host_kdump.sh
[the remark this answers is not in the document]
spmcmillan: https://github.com/canonical/tdx/pull/237/commits/49834af5bd0588025ff659bb50199732297263c7

tests/tests/test_host_kdump.sh
spmcmillan: I changed my mind and made this return -1.

dannf: Note that exit codes are unsigned, so this is the same as exit 255. fwiw, my completely unsolicited opinion is that exiting with the `ssh` error code feels like the typical approach. In fact, I suggest using `set -e` if at all possible.

tests/tests/test_host_kdump.sh
```diff
@@ -0,0 +1,104 @@
...
+# SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+if [ ! -v DEVICE_IP ]; then
+    echo "Must define DEVICE_IP"
+    exit -1
```
hector-cao, line 20: as @dannf pointed out, error code should be positive, returning `-1` will end up as `255`, i think it is better to use exit 1 as `1` is to catch generic errors

tests/tests/test_host_kdump.sh
```diff
@@ -0,0 +1,104 @@
...
+fi
+
+echo "Installing linux-crashdump and rebooting"
+ssh ubuntu@$DEVICE_IP sudo apt install linux-crashdump
+if [ $? != 0 ]; then
+    echo "Can't install linux-crashdump"
+    exit -1
+fi
+
+ssh ubuntu@$DEVICE_IP sudo systemctl reboot
```
hector-cao, line 30: We usually use following options for `ssh` to be more robust on host keys checking:
```
SSH_OPTIONS=" -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"                  
```

hector-cao (dismissed):

tests/tests/test_host_kdump.sh
spmcmillan: Thanks Dan, makes sense

tests/tests/test_host_kdump.sh
```diff
@@ -0,0 +1,78 @@
...
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+# SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+SSH_OPTIONS=" -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
```
frenchwr, line 18: Doesn't this need to get passed to each of the `ssh` commands below (`ssh $SSH_OPTIONS ubuntu@$DEVICE_IP ...`)? I don't think it's a standard shell var that `ssh` reads, is it?

tests/tests/test_host_kdump.sh
frenchwr: Ah I see, so we're expecting the machine to still be in the process of rebooting after 15 seconds (or 10 seconds in other cases below). I imagine this generally works well but I'm thinking about some corner cases in the future involving servers with super fast boot disks or that use some sort of in-memory hibernation for fast rebooting.  

I wonder if testing this with commands that directly track and report reboot details may be more reliable. I Google'd around and [this thread](https://unix.stackexchange.com/questions/747014/command-to-check-if-the-machine-was-rebooted) mentions a few methods (`journalctl --list-boots` and `systemctl show --property=ActiveEnterTimestamp systemd-udevd.service`), although I don't know how they behave on crashes.

tests/tests/test_host_kdump.sh
spmcmillan: dangit, i'll slow down :-)

spmcmillan: https://github.com/canonical/tdx/pull/237/commits/09beade5530cad73a37dc7ebe03bf71647f289a0

frenchwr (approved):
LGTM! I like the changes.

PR #239, opened 2024-10-01
add network to guest to make it accessible via IP  

hector-cao:
This work has been motivated by the demo we were trying to do for Confidential RAG where we will need multiple TD VMs to communicate between them.

With this change, the guest will be accessible via it own IP:

```shell
ssh root@<IP>
```

instead of 

```shell
ssh -p <forward-port> root@localhost
```

And `tdvirsh list` will also print the guest IP:

```
$ ./tdvirsh list
Id   Name                                                        State 
--------------------------------------------------------------------------- 
2    tdvirsh-trust_domain-508f629e-4e86-4969-8517-da1ba35258ed   running (ip:192.168.122.212, hostfwd:44321, cid:3)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,8 @@ TD, PID: 111924, SSH : ssh -p 10022 root@localhost
         1    tdvirsh-trust_domain-f7210c2b-2657-4f30-adf3-639b573ea39f   running (ip:192.168.122.212, hostfwd:32855, cid:3)
         ```
 
-        NOTE: `32855` in `hostfwd:32855` is the port number a user can use to connect to the TD via `ssh`.
+        NOTE: `32855` in `hostfwd:32855` is the port number a user can use to connect to the TD via `ssh -p 32855 root@localhost`.
+              You can also connect to the guest using its IP address : `ssh root@192.168.122.212`.
 
    * A TD can be removed with the following command:
 
--- a/README.md
+++ b/README.md
@@ -243,10 +243,10 @@ TD, PID: 111924, SSH : ssh -p 10022 root@localhost
         ```console
         Id   Name                                                        State
         ---------------------------------------------------------------------------
-        1    tdvirsh-trust_domain-f7210c2b-2657-4f30-adf3-639b573ea39f   running (ssh:32855, cid:3)
+        1    tdvirsh-trust_domain-f7210c2b-2657-4f30-adf3-639b573ea39f   running (ip:192.168.122.212, hostfwd:32855, cid:3)
         ```
 
-        NOTE: `32855` in `ssh:32855` is the port number a user can use to connect to the TD via `ssh`.
+        NOTE: `32855` in `hostfwd:32855` is the port number a user can use to connect to the TD via `ssh`.
 
    * A TD can be removed with the following command:
 
--- a/guest-tools/regular_vm.xml.template
+++ b/guest-tools/regular_vm.xml.template
@@ -36,6 +36,11 @@
       </backingStore>
       <target dev="vda" bus="virtio"/>
     </disk>
+    <interface type='network'>
+      <source network='default'/>
+      <model type='virtio'/>
+      <address type='pci' domain='0x0000' bus='0x06' slot='0x00' function='0x0'/>
+    </interface>
     <console type='pty'>
       <target type='virtio' port='1'/>
     </console>
--- a/guest-tools/tdvirsh
+++ b/guest-tools/tdvirsh
@@ -172,7 +172,11 @@ print_all() {
                     --hmp info qtree 2>&1 |
                     awk '/guest-cid/ {print $3}'
                      )
-            extra_info="(ssh:$host_port, cid:${guest_cid})"
+            host_ip=$(
+               virsh domifaddr ${td_domain}  |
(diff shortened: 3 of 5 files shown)


frenchwr (commented):
Can you still ssh in using localhost + port or does this change mean you can only use the IP? In either case we should update the README to describe what is supported.

About the `tdvirsh list` output:

```
$ ./tdvirsh list
Id   Name                                                        State 
--------------------------------------------------------------------------- 
1    tdvirsh-trust_domain-508f629e-4e86-4969-8517-da1ba35258ed   running (ip:192.168.122.212/24, ssh:45585, cid:3)
```

* I don't think you need to show the CIDR for the IP (the `/24` should be stripped) 
* Now that there is a different method for `ssh` supported (or multiple), I think the `ssh:45585` should be clarified or re-named

hector-cao:
> Can you still ssh in using localhost + port or does this change mean you can only use the IP? In either case we should update the README to describe what is supported.
> 
> About the `tdvirsh list` output:
> 
> ```
> $ ./tdvirsh list
> Id   Name                                                        State 
> --------------------------------------------------------------------------- 
> 1    tdvirsh-trust_domain-508f629e-4e86-4969-8517-da1ba35258ed   running (ip:192.168.122.212/24, ssh:45585, cid:3)
> ```
> 
> * I don't think you need to show the CIDR for the IP (the `/24` should be stripped)
> * Now that there is a different method for `ssh` supported (or multiple), I think the `ssh:45585` should be clarified or re-named
- 1) I just too what virsh gave me, i think we can let it as it is since it will require extra code if we want to remove the `/24` 
- 2) Yes, i will rename the ssh part, that makes me think that i will also need to update the README,

frenchwr:
> I just too what virsh gave me, i think we can let it as it is since it will require extra code if we want to remove the /24

I think it will cause confusion because many users will not understand what it means, and it may lead to issues as you cannot pass a subnet to `ssh`.

hector-cao:
> > I just too what virsh gave me, i think we can let it as it is since it will require extra code if we want to remove the /24
> 
> I think it will cause confusion because many users will not understand what it means, and it may lead to issues as you cannot pass a subnet to `ssh`.

Ok, got it, i will do the change. Thanks !

README.md
```diff
@@ -243,10 +243,10 @@ TD, PID: 111924, SSH : ssh -p 10022 root@localhost
         ```console
         Id   Name                                                        State
         ---------------------------------------------------------------------------
-        1    tdvirsh-trust_domain-f7210c2b-2657-4f30-adf3-639b573ea39f   running (ssh:32855, cid:3)
+        1    tdvirsh-trust_domain-f7210c2b-2657-4f30-adf3-639b573ea39f   running (ip:192.168.122.212, hostfwd:32855, cid:3)
         ```
 
-        NOTE: `32855` in `ssh:32855` is the port number a user can use to connect to the TD via `ssh`.
+        NOTE: `32855` in `hostfwd:32855` is the port number a user can use to connect to the TD via `ssh`.
```
spmcmillan, line 249: I'm confused by this without looking at the rest (will is doing that). 

So, can we connect two ways now?

ssh 192.168.122.212
ssh 127.0.0.1 -p 32855

Or is it:
ssh 192.168.122.212 -p 32855

?

hector-cao: We can connect in two ways, i kept the port forward method for backward comptability

spmcmillan: I would suggest to make it clear in that note in the readme.  Maybe providing two examples of how to run ssh or however you see best.

hector-cao: Thanks for the suggestion but I think that for the sake of the README simplicity, we just need to offer one method, and the method we pick up to suggest is the one that has been here from the beginning and users are familiar with, for now, no need to recommend the other method nor advertise it, if in the future, if there is a need to change, we will do

frenchwr: I think to @spmcmillan 's point I could still see this causing confusion. I would suggest explicitly showing example `ssh` commands that are supported.

Suggested change (I haven't tested these commands so please check):

NOTE: you can `ssh` to the guest in one of two ways: 

* Use the guest IP address directly:

```
ssh root@192.168.122.212
```

* Use a port on the host that is forwarded to the ssh port on the guest:

```
ssh -p 32855 root@localhost
```

frenchwr: Sorry, I was writing that comment while you both were already discussing the same thing. :sweat_smile:

hector-cao: Same answer than the one for @spmcmillan, i do not want to advertise this feature for now because it is for internal need,

frenchwr: I'll go ahead and approve because I think it's your call but I think that because the new feature exposes new information to the user that it's reasonable to update the documentation to explain what that information means.

frenchwr (approved):

README.md
hector-cao: @spmcmillan @frenchwr I finally made the change, i think if you both find it confusing, others might find it confusing as well. Please take a look if it works for you : https://github.com/canonical/tdx/pull/239/commits/3011929e8b1c1c04687a354f27c27326e00c2951

frenchwr: Looks good! Thanks. :)

PR #245, opened 2024-10-10
fix failing tests cause of permanent modification of qemu machine arguments

hector-cao:
currently in the test tests/test_quote_configfs_tsm.py::test_qgs_socket we modify permanently the qemu -machine and -object arguments. this can make following tests fail

we do not want to modify these arguments permanently but only for the current test

### DIFF
--- a/tests/tests/test_quote_configfs_tsm.py
+++ b/tests/tests/test_quote_configfs_tsm.py
@@ -41,8 +41,8 @@ def test_qgs_socket(qm):
     """
     Test QGS socket (No Intel Case ID)
     """
-    object = '{"qom-type":"tdx-guest","id":"tdx","quote-generation-socket":{"type": "vsock", "cid":"2","port":"4050"}}'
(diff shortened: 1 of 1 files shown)


mckees:
I'm seeing that this is a more precise way to enable the needed socket, but where does this end up getting changed back? Or are you just altering a temporary state that gets destroyed when the test gets cleaned up?

hector-cao:
> I'm seeing that this is a more precise way to enable the needed socket, but where does this end up getting changed back? Or are you just altering a temporary state that gets destroyed when the test gets cleaned up?

Yes, this change is done in the context of the `qm` object (passed as argument here : https://github.com/canonical/tdx/blob/b3c9ee0e0bffc59832263beaf1a2ba173e2ab37a/tests/tests/test_quote_configfs_tsm.py#L40), this object will be destroyed when we exit the test function (`test_qgs_socket`). With the old source code, the change still persists after the destruction of the object `qm`

mckees (approved):
Alright, LGTM then

PR #249, opened 2024-10-20
fix failing test vsock in checkbox provider

hector-cao:
vsock test uses a customized iperf tool with vsock support
this tool is installed on the host

however, checkbox24 comes with iperf tool and it will be used by the test,
without vsock support, the test fails

we device to bundle the customized iperf tool to the checkbox provider
and set PATH variable to use the iperf in the provider

one other solution would be to set the PATH to use the iperf installed on the system
but we think that bundle iperf3 into the provider is a better approach


mckees (approved):
LGTM!

PR #251, opened 2024-10-22
tests : add stress test for quote generation

hector-cao:
manually, on some systems, when we loop on requesting the quote generation, the host seems to be in heavy load and becomes unresponsive, add the stress test case for that.


frenchwr (approved):
LGTM

PR #272, opened 2024-11-14
Noble 24.04 minor improvements

hector-cao:
This PR comes with 2 commits:

- output the qemu command to a script
this will allow users to run manually the guest after the test
this capability is useful for debugging purposes
for example, the user can run the guest manually by doing:
```bash
$ sudo /tmp/tdxtest-default-hk8u6mf3/run.sh
```
`hk8u6mf3` is random and may vary for each test

- right now, we use `proc.terminate()` to send SIGTERM to qemu process
this tends to corrupt the rootfs
this will prevent us from running the guest again with this rootfs
try to stop the VM (shutdown) properly first and terminate() it
later if necessary

### DIFF
--- a/tests/lib/Qemu.py
+++ b/tests/lib/Qemu.py
@@ -352,6 +352,9 @@ class QemuMonitor():
     def wakeup(self):
         self.send_command("system_wakeup")
 
+    def powerdown(self):
+        self.send_command("system_powerdown")
+
     def __del__(self):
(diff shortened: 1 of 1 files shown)


frenchwr (approved):
LGTM. Only thing I'm wondering is if you need or want to document how to use the qemu troubleshooting script, but I leave it up to you, so not a blocker.

PR #278, opened 2024-11-20
add the management of multiple ubuntu releases

hector-cao:
for now, this branch only supports 24.04, with the upcoming 24.10 support, we would like this branch to offer it instead of using a different branch

this helps us to lower the maintenance effort since the majority of the code base is release agnostic. (especially tests, tools, documentation ....)

### DIFF
--- a/guest-tools/image/create-td-image.sh
+++ b/guest-tools/image/create-td-image.sh
@@ -43,13 +43,15 @@ fi
 
 LOGFILE=/tmp/tdx-guest-setup.txt
 FORCE_RECREATE=false
-OFFICIAL_UBUNTU_IMAGE=${OFFICIAL_UBUNTU_IMAGE:-"https://cloud-images.ubuntu.com/releases/noble/release/"}
-CLOUD_IMG=${CLOUD_IMG:-"ubuntu-24.04-server-cloudimg-amd64.img"}
+CODE_NAME=$(lsb_release -cs)
+UBUNTU_VERSION=$(lsb_release -rs)
+OFFICIAL_UBUNTU_IMAGE=${OFFICIAL_UBUNTU_IMAGE:-"https://cloud-images.ubuntu.com/releases/${CODE_NAME}/release/"}
+CLOUD_IMG=${CLOUD_IMG:-"ubuntu-${UBUNTU_VERSION}-server-cloudimg-amd64.img"}
 CLOUD_IMG_PATH=$(realpath "${SCRIPT_DIR}/${CLOUD_IMG}")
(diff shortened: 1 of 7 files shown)


mckees (approved):
LGTM

guest-tools/image/create-td-image.sh
```diff
@@ -43,13 +43,15 @@ fi
 
 LOGFILE=/tmp/tdx-guest-setup.txt
 FORCE_RECREATE=false
-OFFICIAL_UBUNTU_IMAGE=${OFFICIAL_UBUNTU_IMAGE:-"https://cloud-images.ubuntu.com/releases/noble/release/"}
-CLOUD_IMG=${CLOUD_IMG:-"ubuntu-24.04-server-cloudimg-amd64.img"}
+CODE_NAME=$(lsb_release -cs)
+UBUNTU_VERSION=$(lsb_release -rs)
+OFFICIAL_UBUNTU_IMAGE=${OFFICIAL_UBUNTU_IMAGE:-"https://cloud-images.ubuntu.com/releases/${CODE_NAME}/release/"}
```
frenchwr, line 48: So for now it's only possible to create a guest image that matches the host OS, correct? I thought you mentioned wanting to support other guest OS's than may differ from the host OS? Is that something that will be enabled later?

setup-tdx-common
```diff
@@ -40,7 +39,7 @@ EOF
 
   cat <<EOF | tee /etc/apt/apt.conf.d/99unattended-upgrades-kobuk-${ppa_id}
 Unattended-Upgrade::Allowed-Origins {
-  "${distro_id}:${distro_codename}";
+  "${distro_id}:${UBUNTU_CODENAME}";
```
frenchwr, line 42: I'm wondering if we should add `UBUNTU_CODENAME=$(lsb_release -cs)` to this script? I know it gets sourced by other scripts that are responsible for setting this variable, but it feels like that could be easily overlooked or prone to errors in the future. Or maybe just include a quick check to futureproof it with something like:

```bash
if [[ -z "${UBUNTU_CODENAME}" ]]; then
  >&2 echo "Please set UBUNTU_CODENAME prior to sourcing setup-tdx-common." 
  exit 1
fi
```

frenchwr (commented):
A few small comments.

setup-tdx-common
hector-cao: You are right, let me propose something different

hector-cao: @frenchwr i would like to propose that 0d3c308a7b3ac35f1f30bafe2846644399827005 , let me what you thiink

frenchwr (approved):
Looks good!

PR #280, opened 2024-11-20
tests : extend ssh connection timeout for test tsc_deadline disable

hector-cao:
on 24.10, the VMs takes > 75s to boot if -tsc-deadline is added to cpu argument of qemu
for now, extend the ssh connection timeout to make the test pass


mckees (approved):
Makes sense, LGTM

PR #282, opened 2024-11-21
image : add multi release image generation capability

hector-cao:
the image generation script will now be able to generate image for multiple ubuntu releases

To generate the image for 24.04:
```
sudo ./create-td-image.sh  -v 24.04
```

The user needs to specify the version, if not, the script will fail, it will not try to guess the version from the ubuntu host since users can run on a ubuntu release we do not offer TDX support for (for example : 22.04)

### DIFF
--- a/guest-tools/image/create-td-image.sh
+++ b/guest-tools/image/create-td-image.sh
@@ -39,6 +39,7 @@
 
 # CLOUD_IMG: the ubuntu image name
 # OFFICIAL_UBUNTU_IMAGE: the url where we can download the CLOUD_IMAGE file
+# UBUNTU_VERSION: the ubuntu version (24.04, 24.10, ...)
 # GUEST_USER: the username in the image
 # GUEST_PASSWORD: the user password in the image
 # GUEST_HOSTNAME: the guest hostname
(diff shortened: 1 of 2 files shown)


frenchwr (approved):
Nice! Approving but just wondering if you want to add documentation to the README or do you prefer keeping it lesser known for now? I'm fine either way.

hector-cao:
> Nice! Approving but just wondering if you want to add documentation to the README or do you prefer keeping it lesser known for now? I'm fine either way.

Thanks, i just added a new commit 093e5f565323a8b114dc6334dfc0a74a71b448f3, can you check and tell me what you think ? (I will request your review again).

For the README, yes, we need to rework it and it will come in the following PR

frenchwr (approved):
So if I'm understanding correctly the user is now required to provide the OS version? If so I would suggest at least updating some of the text and the example in [this section](https://github.com/canonical/tdx?tab=readme-ov-file#5-create-td-image). Also fine if you plan to do that in a quick follow up PR.

hector-cao:
> So if I'm understanding correctly the user is now required to provide the OS version? If so I would suggest at least updating some of the text and the example in [this section](https://github.com/canonical/tdx?tab=readme-ov-file#5-create-td-image). Also fine if you plan to do that in a quick follow up PR.

Yes, i will do a follow up for the README since i want to involve @bktan8 into its review

PR #288, opened 2024-11-26
Improve Supported Hardware section

bktan8:


hector-cao (changes requested):
The rendering is not good at my side

image

bktan8:
> The rendering is not good at my side
> 
> image

Ooops!  Missed a | in the table formatting.  Fixed.

hector-cao (approved):
Awesome ! +1

PR #295, opened 2024-12-04
enable normal user for libvirt in setup script

hector-cao:
enable by default non root user for libvirt
since everytime, I run into permissions issue and realized I missed this step in the README to enable non-root user
It is a waste of time and this commit tries to automate this for the users

### DIFF
--- a/setup-tdx-common
+++ b/setup-tdx-common
@@ -98,3 +98,27 @@ grub_set_kernel() {
     fi
     grub_switch_kernel "${KERNEL_RELEASE}"
 }
+
+# Enable non-root user to run TDs
+# Try to detect the normal user when the script is run with sudo
+# If the script is run as root, just skip this step
+#
+# This script modifies the configuration file /etc/libvirt/qemu.conf:
+# user = <user>
+# group = <group>
+# dynamic_ownership = 0
+enable_normal_user_libvirt() {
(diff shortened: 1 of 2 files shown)


frenchwr (approved):
Nice. I'm approving but I'm also wondering if it would be better to instead run from `tdvirsh`...i.e. check if the user has permissions and if not run `enable_normal_user_libvirt`. I think where you have it probably makes more sense as it's a one-time check. I'm curious what you think?

hector-cao:
> Nice. I'm approving but I'm also wondering if it would be better to instead run from `tdvirsh`...i.e. check if the user has permissions and if not run `enable_normal_user_libvirt`. I think where you have it probably makes more sense as it's a one-time check. I'm curious what you think?

I like very much the concept of lazy initialization generally, for this particular use case we are talking about here, the action we need to take will make libvirtd restart (and par consequence stop all existing VMs if any), I m also fine if we do what you suggest for the permission detection but I propose to do the modification automatically but suggest the user to run the script. I did this MR because I m frustrated to spend every time some effort to realize that I have to modify the libvirt conf however, I m not completely sure if I do the right thing, I m afraid of modifying this conf automatically for all users because there might be some who do not want that ...

hector-cao:
@frenchwr What do you think ?

PR #296, opened 2024-12-05
add direct boot methods and UKI

hector-cao:
we have different methods to boot a TD
for now, we use the traditional method that consists of passing
the ubuntu image to qemu
however, we have other ways to boot a TD:
- direct boot : by providing the kernel and initrd to qemu
- UKI : by providing the Unified Kernel image (that contains the kernel and initrd)

each method generates a different event log journal because the boot chain
involves less or more components that generate event logs

it is important to evaluate which is the best boot method to have a short
and consistent event log journal to make sure that remote attestation
is reliable

Please also check: https://github.com/canonical/tdx/blob/noble-24.04-direct-boot/guest-tools/direct-boot/README.md

PR #299, opened 2024-12-10
improvements on the direct boot using UKI

hector-cao:
the improvements are results of the user's feedback on the closed MR : https://github.com/canonical/tdx/pull/296

+ grammar check on README

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,9 @@ Two Ubuntu releases are currently supported for base host OS and guest OS:
 * Ubuntu Noble 24.04 LTS
 * Ubuntu Oracular 24.10
 
-Follow these instructions to setup the Intel TDX host, create a TD, boot the TD, and attest the integrity of the TD's execution environment.
+Follow these instructions to set up the Intel TDX host, create a TD, boot the TD, and attest the integrity of the TD's execution environment.
 
-The host OS and TD setup can be cutomized by editing the global configuration file: `setup-tdx-config`.
+The host OS and TD setup can be customized by editing the global configuration file: `setup-tdx-config`.
 
 <a id="report-an-issue"></a>
 ## 2. Report an Issue
@@ -77,7 +77,7 @@ Download and install appropriate Ubuntu Server on the host machine:
 
 2. Customize the setup of the host and TD by editing the configuration file `setup-tdx-config`.  
     By default, remote attestation components are not installed on the host and inside the TD.  
-    You can chose to automatically install remote attestation packages provided by Canonical by setting `TDX_SETUP_ATTESTATION=1`.  
+    You can choose to automatically install remote attestation packages provided by Canonical by setting `TDX_SETUP_ATTESTATION=1`.  
    In this case, you can skip [step 8.2.1](#step-8-2-1) and [step 8.3.3](#step-8-3-3).  
 
 3. Run the `setup-tdx-host.sh` script.<a id="step-4-2-3"></a>
@@ -156,7 +156,7 @@ You can pass `24.04` to the `-v` to generate a TD image based on Ubuntu 24.04.
 
 The resulting image will be based on an ([`Ubuntu cloud image`](https://cloud-images.ubuntu.com/)),
 the default root password is `123456`, and other default settings are used.
-Please note the most important options described after the commands and have a look at the `create-td-image.sh` script for more available options.
+Please note the most important options described after the commands and take a look at the `create-td-image.sh` script for more available options.
 
 Important options for TD image creation:
 * If you're behind a proxy, use `sudo -E` to preserve user environment.
@@ -385,7 +385,7 @@ sudo ./check-production.sh
 	```bash
 	sudo systemctl status pccs
 	```
-6. To setup the PCCS in the next step, you need a subscription key for the [Intel PCS](https://api.portal.trustedservices.intel.com/provisioning-certification).
+6. To set up the PCCS in the next step, you need a subscription key for the [Intel PCS](https://api.portal.trustedservices.intel.com/provisioning-certification).
    1. If you did not request such a subscription key before, [subscribe](https://api.portal.trustedservices.intel.com/products#product=liv-intel-software-guard-extensions-provisioning-certification-service) 
       to Intel PCS, which requires to log in (or create an account). Two subscription keys are generated (for key rotation) and both can be used for the following step.
    2. If you did request such a subscription key before, [retrieve](https://api.portal.trustedservices.intel.com/manage-subscriptions) one of your keys, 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -593,7 +593,8 @@ you proceed to [step 4](#verify-itts-client-version).
 
 One of the key components remote attestation is based on is the runtime measurement values (stored in RTMRs
 registers for each TD by the TDX module). These RTMR values are computed from the digests of the entries of
-the TD boot event log.
+the TD boot event log. During the system's boot, each event's digest (SHA384) is extended (appended) to the
+current value of the associated RTMR. The resulting value's digest becomes the new value of the RTMR.
```
mythi, line 597: the event type determines what's in the hash so it's not automatically  'digest of the event'.

hector-cao:
> LGTM, thanks! Nice to see `-kernel` working easily for you as well

Thanks for the feedback that greatly improved the instructions and script

README.md
```diff
@@ -593,7 +593,8 @@ you proceed to [step 4](#verify-itts-client-version).
 
 One of the key components remote attestation is based on is the runtime measurement values (stored in RTMRs
 registers for each TD by the TDX module). These RTMR values are computed from the digests of the entries of
-the TD boot event log.
+the TD boot event log. During the system's boot, each event's digest (SHA384) is extended (appended) to the
```
mckees: If you're using "appended" to clarify the meaning of "extended", maybe we should just go with "appended" unless "extended" is baked into the language of this component

guest-tools/direct-boot/boot_uki.sh
```diff
@@ -46,15 +46,13 @@ EOM
 
 cleanup() {
     echo "cleanup ..."
```
mckees, line 48: If you're deleting the functional part of the cleanup, is there a reason you're keeping the function that just echoes?

README.md
```diff
@@ -591,10 +591,10 @@ you proceed to [step 4](#verify-itts-client-version).
 
 #### 9.1. Event log and measurements
 
-One of the key components remote attestation is based on is the runtime measurement values (stored in RTMRs
-registers for each TD by the TDX module). These RTMR values are computed from the digests of the entries of
-the TD boot event log. During the system's boot, each event's digest (SHA384) is extended (appended) to the
-current value of the associated RTMR. The resulting value's digest becomes the new value of the RTMR.
+One of the key components remote attestation is the runtime measurement values (stored in RTMRs
```
mckees, line 594: Add "of" after "componenets"

mckees: "One of" implies a singular subject, but "values" is plural. Might be better to just flip the sentence around and say something like, "Runtime measurement values (paren contents) are key components of remote attestation."

README.md
```diff
@@ -591,10 +591,10 @@ you proceed to [step 4](#verify-itts-client-version).
...
+registers for each TD by the TDX module). During the system boot, each component of the boot process (binary or conf)
+is measured into a digest. This digest value is extended (appended) to the RTMR's current value. The digest of the result value
```
mckees, line 596: Same thing with appended/extended. 

Also, I'm not sure what is meant by "measured into a digest". I think what you're meaning here is that each component of the boot process becomes input for the digest/hash function?

guest-tools/direct-boot/boot_uki.sh
hector-cao: thanks, i will remove it

README.md
hector-cao: Done 57bef18762575d7cfbabc876ec221d46df674c1e

guest-tools/direct-boot/boot_uki.sh
hector-cao: Done 57bef18762575d7cfbabc876ec221d46df674c1e

README.md
hector-cao: Done 57bef18762575d7cfbabc876ec221d46df674c1e

README.md
hector-cao: digest is the value from the hash function that is applied to data that needs to be measured

qinkunbao:
Hi Hector,

I've been reviewing the TPM measurement sequence during direct boot and noticed a potential discrepancy that could have security implications.  I'm seeing conflicting behavior between how OVMF and the EFI handle measurements for RTMR and vTPM. I just wondered if you have checked the event log of vTPM and RTMR on a TDX CVM using an image that supports the direct boot.

OVMF (https://github.com/tianocore/edk2/blob/62de957185d71feb9c4d09f4eddcdac1980632fb/SecurityPkg/Library/DxeTpm2MeasureBootLib/DxeTpm2MeasureBootLib.c#L478) attempts to measure into RTMR first and skips vTPM if the RTMR measurement succeeds.

On the other hand, the EFI/libstub, as implemented in the Linux kernel patches (https://patches.linaro.org/project/linux-efi/list/?series=236617), prioritizes measurement into vTPM and only proceeds to RTMR if the vTPM measurement fails.

This conflicting approach results in an incomplete TPM and RTMR event log on systems using direct boot.  Because either OVMF or EFI skips its part of the measurement, neither RTMR nor vTPM has a complete log. This incomplete log is a security concern as it undermines the integrity and verifiability of the boot process.

mythi:
(the TCG vs CC measurement protocol topic could be a separate issue)

> This conflicting approach results in an incomplete TPM and RTMR event log on systems using direct boot.

I think the general consensus still is that only one measurement protocol is available at a time.

qinkunbao:
> (the TCG vs CC measurement protocol topic could be a separate issue)
Yes, I agree. I just wanted to say the direct boot doesn't work for TDX CVMs.

> I think the general consensus still is that only one measurement protocol is available at a time.
The firmware should enforce that only only one of the protocol is used. It is a big security issue. The boot chain software should measure into both to avoid the security issue such as https://msrc.microsoft.com/update-guide/vulnerability/CVE-2021-42299

mythi:
> Yes, I agree. I just wanted to say the direct boot doesn't work for TDX CVMs.

Just to clarify: you mean that the installed guest kernel version is too low / Ubuntu kernel does not have the necessary EFI/libstub patches for RTMR measurements?

README.md
```diff
@@ -591,10 +591,10 @@ you proceed to [step 4](#verify-itts-client-version).
 
 #### 9.1. Event log and measurements
 
-One of the key components remote attestation is the runtime measurement values (stored in RTMRs
-registers for each TD by the TDX module). During the system boot, each component of the boot process (binary or conf)
-is measured into a digest. This digest value is extended (appended) to the RTMR's current value. The digest of the result value
-becomes the new value of the RTMR.
+One of the key components the remote attestation is the runtime measurement. The runtime measurement values
```
mckees, line 594: I still think this should be "One of the key components of remote attestation is the runtime measurement"

hector-cao: Thanks, done : 54db08b0664d20501139609d6f095e2fe601e2db

mckees (approved):
LGTM, thanks!

PR #304, opened 2024-12-16
tests : skip attestation tests if platform is not registered

hector-cao:
when the tests run on platform that is not registered to the intel trust service, quote cannot be generated. in that case, the tests must be skipped.

this commit add a script `check-registration.sh` that allows to detect if the platform has been properly registered, if not, all the tests marked as quote_generation will be skipped

### DIFF
--- a/attestation/check-registration.sh
+++ b/attestation/check-registration.sh
@@ -47,7 +47,7 @@ then
     then
         check_mpa_status
     else
-        echo "mpa_manage is available, assuming platform not registered."
+        echo "mpa_manage is not available, assuming platform not registered."
     fi
 fi
 
--- a/attestation/check-registration.sh
+++ b/attestation/check-registration.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+# This file is part of Canonical's TDX repository which includes tools
+# to setup and configure a confidential computing environment
+# based on Intel TDX technology.
+# See the LICENSE file in the repository for the license text.
+
+# Copyright 2024 Canonical Ltd.
+# SPDX-License-Identifier: GPL-3.0-only
+
+# This program is free software: you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 3,
+# as published by the Free Software Foundation.
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranties
+# of MERCHANTABILITY, SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU General Public License for more details.
+
+if [ "$EUID" -ne 0 ]
+  then echo "Please run as root"
+  exit
+fi
+
+# check platform registration
+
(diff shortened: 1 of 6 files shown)


attestation/check-registration.sh
```diff
@@ -0,0 +1,49 @@
...
+    # If the platform has been registered successfully, the command will output:
+    # Last reported registration error code: 0
+    if mpa_manage -get_last_registration_error_code | grep "error code: 0" 2>&1 > /dev/null
+    then
+	REGISTRATION_SUCCESS=0
+    fi
+}
+
+# check if MPA is used for platform registration
+if systemctl is-enabled mpa_registration_tool.service  2>&1 > /dev/null
```
frenchwr, line 41: TIL that `systemctl is-enabled` is a thing - nice!

attestation/check-registration.sh
```diff
@@ -0,0 +1,49 @@
...
+then
+    if command -v mpa_manage 2>&1 > /dev/null
```
frenchwr, line 43: I assume `mpa_manage` is not installed by default - does this come with a package from the attestation PPA or somewhere else? I'm wondering if it's worthwhile to add an else that logs an error rather than just silently skipping `check_mpa_status`.

tests/tests/conftest.py
```diff
@@ -1,9 +1,29 @@
...
 import util
 
+script_path=os.path.dirname(os.path.realpath(__file__))
+
+# Is platform registered for quote generation
+def is_platform_registered():
+    try:
+        subprocess.check_call([f'{script_path}/../../attestation/check-registration.sh'])
+    except:
+        return 0
```
frenchwr, line 15: Can you use `False` and `True` here? This confused me at first because I saw it was a shell script being called and therefore assumed zero should mean success and non-zero should mean failure.

frenchwr (commented):
Left a few comments

attestation/check-registration.sh
hector-cao:
```
$ dpkg -S /usr/bin/mpa_manage
sgx-ra-service: /usr/bin/mpa_manage
```

Agree for the log, let me do that

hector-cao: Done : ac138eb2cb521a414d23b294146a51da6fac35a0

tests/tests/conftest.py
hector-cao: Done : 9eee5b3dd53d405d6c8923f37771668827b6d5b6

attestation/check-registration.sh
frenchwr: @hector-cao In the echo command you added I think you forgot "not", "mpa_manage is not available, ..."

hector-cao: Oups, thanks for the catch, fixed : c8b0152811c36b8b8b00228997607c7606f8e9c4

frenchwr (approved):
LGTM!

PR #306, opened 2024-12-17
fix registration status check

hector-cao:
mpa_manage -get_last_registration_error_code is not the most accurate mean to check the registration status because it outputs stauts 0 when the registration is in progress (not done)


frenchwr (approved):
LGTM!

PR #307, opened 2025-01-06
tests reorganization and add main test script

hector-cao:
1 - reorganization of the tests folder
    tests having same category live in a subfolder

2 - expose pytest capabilities to users
    currently, tox is masking all pytest features and only
    exposes what we specify in tox.ini file
    since tox is only here to manage the test environment
    it should be transparent to the users for tests execution
    and management. the users should be able to take full advantage
    of what pytest offers.
    we introduced a new script tdtest that hides tox and exposes
    all pytest features to users


tests/README.md
```diff
@@ -4,88 +4,30 @@ This folder contains Intel TDX tests.
...
+Please run `tdtest -h` for more details about its usage.
 
-```
-$ sudo -E tox -e test_host
-```
+The tests are organized in different categories and this organization is
+reflected in the structure of the `tests`.
 
-- Run guest tests:
+You can choose to run a category of tests by specifying the folder:
```
frenchwr, line 17:
```suggestion
You can choose to run a category of tests by specifying the appropriate sudirectory under `tests`. For example, to run the boot tests:
```

tests/tdtest
```diff
@@ -0,0 +1,81 @@
...
+  To list tests in boot folder:
+    $ sudo $(basename "$0") --co tests/boot
+
+---
+EOM
+}
+
+parse_params() {
+    # program cannot be run without argument
+    #[[ -z "$@" ]] && usage && exit 0
```
frenchwr, line 35: Should this be uncommented?

frenchwr (commented):
Few minor comments. Otherwise looks great.

tests/tdtest
hector-cao: no, in fact, the script can be invoked without argument to run all the tests, i removed the commented lines at : 3ac68f3e075ae47d8d1b26e5b5c4d56c37ff7229

frenchwr (approved):
LGTM!

PR #309, opened 2025-01-10
README.md : add needed package for kernel build

hector-cao:


frenchwr (approved):
LGTM but probably @ThibF should also take a look.

ThibF (approved):

PR #322, opened 2025-01-29
Export image path for direct boot

jkr0103:
`direct-boot` scripts is missing the functionality to use images built with non-generic kernel (e.g. with an intel kernel). To use the non-generic kernel images, user have to modify the scripts.

This PR eliminates the need to modify the script instead, image name can we provided with cmdline while invoking the script.

### DIFF
--- a/guest-tools/direct-boot/README.md
+++ b/guest-tools/direct-boot/README.md
@@ -55,9 +55,20 @@ This script will generate 3 files:
 
 ### Direct boot
 
+Boot TD with the provided script.
+By default, the script will use an image with a generic kernel located at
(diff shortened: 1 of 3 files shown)


hector-cao:
@jkr0103  for direct boot, we need 3 artifacts : kernel, initrd and the image, for now, we do not want to offer to user the option to use a customized version of 1 of those 3 artifacts because I think they might be tightly couples (kernel modules in the image have to match the kernel ... etc)

hector-cao:
@jkr0103 I will close this MR, feel free to open it again if you feel this feature is a must-have

PR #329, opened 2025-02-21
Clarify default behavior of run-td.sh script

bktan8:


README.md
```diff
@@ -195,9 +195,11 @@ The script provided for the QEMU method supports running only a single instance.
 ### 6.1 Boot TD with QEMU
 
 Boot TD with the provided script.
-By default, the script will use an image with a generic kernel located at
-`./image/tdx-guest-ubuntu-<24.04|24.10>-generic.qcow2`. A different qcow2
-image (e.g., one with an intel kernel) can be used by setting the `TD_IMG`
+If no parameter is provided, the script will use the generic kernel image
```
hector-cao, line 198: Can we move this to the section 6 since this also applies to `6.2`

README.md
```diff
@@ -195,9 +195,11 @@ The script provided for the QEMU method supports running only a single instance.
...
 
 Boot TD with the provided script.
-By default, the script will use an image with a generic kernel located at
-`./image/tdx-guest-ubuntu-<24.04|24.10>-generic.qcow2`. A different qcow2
-image (e.g., one with an intel kernel) can be used by setting the `TD_IMG`
+If no parameter is provided, the script will use the generic kernel image 
+with the same Ubuntu version as the host. For example, if you're running 
+Ubuntu 24.04 host, the default image will be 
+`tdx-guest-ubuntu-24.04-generic.qcow2`. A different qcow2
+image (e.g. one with an intel kernel) can be used by setting the `TD_IMG`
```
hector-cao, line 202: or using the argument `--td-image` of the `tdvirsh` command

hector-cao (changes requested):
:wave:

README.md
bktan8: I decided to repeat the statement in 6.1 and 6.2.  I found it easier to understand when it's together with the "different qcow2" statement.

README.md
```diff
@@ -210,7 +213,11 @@ cd tdx/guest-tools
 An example output:
 
 ```bash
-TD, PID: 111924, SSH : ssh -p 10022 root@localhost
+TD started by QEMU with PID: 13392.
```
hector-cao, line 216: @bktan8 I do not understand this line, can you help me please

bktan8: The TD has started by QEMU and the resulting PID is xyz.  I updated the text.  Hope it's clearer.

hector-cao: why we would want to mention the TD PID ?

bktan8: I was just trying to keep what was done before, which was "TD, PID: 111924, SSH : ssh -p 10022 root@localhost".  It's just information.  I'm fine remove the PID info if that's your preference.

hector-cao: I did let this PID stuff in on one of the PR, i should have not, so either we keep it as it is or we remove it but no need to transform / modify it

bktan8: So something like?  

```
TD started by QEMU.
To log in with the non-root user (default: tdx / password: 123456), as specified in setup-tdx-config, use:
   $ ssh -p 10022 <username>@localhost
To log in as root (default password: 123456), use:
   $ ssh -p 10022 root@localhost
```

bktan8: ok.  Back to what it was.

hector-cao (approved):
lgtm :rocket:

PR #334, opened 2025-03-11
Make grub-install to not alter the EFI boot order

hector-cao:
Right now, when we run the TDX setup, we call grub-install to finalize some kernel boot parameters additions, grub-install modifies the EFI boot order (by modifying the NVRAM). This boot order change will place Ubuntu as the first boot entry and breaks by the way MAAS that expects the machine to PXE boot in the first place.

We use `--no-nvram` argument to tell grub-install to not change the boot order


ThibF (approved):
Great finding! didn't realize this could be an issue.

PR #339, opened 2025-03-20
Add SGX_DEBUG_MODE check for attestation

fanchen2:


attestation/check-production.sh
```diff
@@ -27,6 +27,7 @@ set -e
 
 modprobe msr
 PROD=$(rdmsr 0xce -f 27:27)
+DEBUG=$(rdmsr 0x503)
```
hector-cao, line 30:
```suggestion
SGX_DEBUG_MODE=$(rdmsr 0x503)
```

attestation/check-production.sh
```diff
@@ -51,3 +52,9 @@ if [ "${PROD}" = "0" ]; then
 else
     echo "Pre-production"
 fi
+
+if [ "${DEBUG}" = "0" ]; then
```
hector-cao, line 56:
```suggestion
if [ "${SGX_DEBUG_MODE}" = "0" ]; then
```

hector-cao (changes requested):
thanks @fanchen2 !

attestation/check-production.sh
```diff
@@ -51,3 +52,9 @@ if [ "${PROD}" = "0" ]; then
 else
     echo "Pre-production"
 fi
+
```
hector-cao, line 55:
```suggestion
printf "SGX_DEBUG_MODE=0x%s\n" ${SGX_DEBUG_MODE}
```

hector-cao (changes requested):

attestation/check-production.sh
fanchen2: How about only print the bit value when it is in debug mode? MR has been updated like below.
```
if [ "${SGX_DEBUG_MODE}" = "0" ]; then
    echo "SGX production mode"
else
    printf "SGX debug unlock mode, SGX_DEBUG_MODE=0x%s\n" ${SGX_DEBUG_MODE}
fi
```

hector-cao: SGX_DEBUG_MODE hols the value of the register, SGX production mode is only 1 bit in this value, so I think the check `[ "${SGX_DEBUG_MODE}" = "0" ]` is not accurate, I propose to only print the value of the register by doing `printf "SGX_DEBUG_MODE=0x%s\n" ${SGX_DEBUG_MODE}` instead of using the if statement

hector-cao (changes requested):

attestation/check-production.sh
fanchen2: updated, thanks

hector-cao (approved):
lgtm, thanks :rocket:

albertoblue87-netizen (commented):
Hecho Sr.

attestation/check-production.sh
albertoblue87-netizen: 🙏🏻

albertoblue87-netizen (commented):
Hecho

PR #345, opened 2025-04-09
add support for PPA outside of kobuk team

hector-cao:


frenchwr (approved):

setup-tdx-guest.sh
```diff
@@ -56,7 +56,7 @@ apt update
 apt install --yes software-properties-common gawk &> /dev/null
 
 # cleanup
-rm -f /etc/apt/preferences.d/kobuk-team-tdx-*
+rm -f /etc/apt/preferences.d/*tdx-*
```
BFuhry, line 59: I now that this PR is already merged and I understand why you removed "kobuk-team-". But "*tdx*" is very generic and might remove other repos added by the customer that coincidentally have tdx in its name.
Same issue in setup-tdx-host.sh.

hector-cao: does it work for you if I add kobuk prefix on top of what we have right now ? `rm -f /etc/apt/preferences.d/kobuk*tdx-*`, i m trying to handle all the backward compatibility with the naming we have for older releases so far

BFuhry: Sure, as soon as you add kobuk, the likelihood of a false match gets far smaller. But as this PR is about "add support for PPA outside of kobuk team", I'm not sure if you still reach your goal with this change.

What is the reason for this cleanup anyhow? Where are
```
rm -f /etc/apt/preferences.d/*tdx-*
rm -f /etc/apt/apt.conf.d/99unattended-upgrades-kobuk
```
added that you remove here?

Somewhat related comment: Why is the function still called add_kobuk_ppa, when it now is more generic?

PR #346, opened 2025-04-15
check-production.sh : add sierra forest CPU model

hector-cao:


attestation/check-production.sh
```diff
@@ -42,6 +42,9 @@ fi
 if [ "$CPU_MODEL" = "175" ]; then
     CPU_GEN="Intel® Xeon® 6 with E-cores (codename: Sierra Forest)"
 fi
+if [ "$CPU_MODEL" = "175" ]; then
+    CPU_GEN="Intel® Xeon® 6 with E-cores (codename: Sierra Forest)"
+fi
```
BFuhry, line 47: Exactly this content is already present in the three lines above.

PR #356, opened 2025-05-19
Fix upgrade for the kernel

hector-cao:
As of now, the setup scripts tells grub (grubenv)
to boot on the latest -intel kernel revision we have at the time of running the script.
If apt upgrade is invoked later and a new revision of -intel is available, this new kernel will not be set as kernel for next boot because apt upgrade is not aware about the setting we did with grubenv. As consequence, the old kernel revision is still the one the host will boot on.

To allow user to upgrade the kernel, this commit adds the `upgrade` option to the setup scripts.

### DIFF
--- a/setup-tdx-guest.sh
+++ b/setup-tdx-guest.sh
@@ -28,6 +28,26 @@ _on_error() {
 }
 trap '_on_error $?' ERR
 
+parse_params() {
+    while :; do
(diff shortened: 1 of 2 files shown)


mckees (approved):
LGTM, but I might suggest some consolidation of the duplicated code to ease maintenance in the future

PR #357, opened 2025-05-19
Add kobuk prefix to apt configuration files

hector-cao:
To pin TDX packages, we add apt configuration files for all the PPAs we use to pull in the packages to the OS. Add the prefix kobuk to these configuration file names to avoid any potential conflict with other TDX setup from Intel or other users.

### DIFF
--- a/setup-tdx-guest.sh
+++ b/setup-tdx-guest.sh
@@ -56,7 +56,10 @@ apt update
 apt install --yes software-properties-common gawk &> /dev/null
 
 # cleanup
-rm -f /etc/apt/preferences.d/*tdx-*
+# NB: '*' before kobuk to keep backward compatiblity to make sure
+# we clean up all conf files that have been deployed in the
+# previous releases
+rm -f /etc/apt/preferences.d/*kobuk*tdx-*
 rm -f /etc/apt/apt.conf.d/99unattended-upgrades-kobuk
 
 # We want wordsplitting if there is multiple entries
--- a/setup-tdx-common
+++ b/setup-tdx-common
@@ -44,7 +44,7 @@ add_kobuk_ppa() {
 
   add-apt-repository -y ppa:${team}/${ppa_id}
 
-  cat <<EOF | tee /etc/apt/preferences.d/tdx-${team}-${ppa_id}-pin-4000
+  cat <<EOF | tee /etc/apt/preferences.d/kobuk-tdx-${team}-${ppa_id}-pin-4000
 Package: *
 Pin: release o=${distro_id}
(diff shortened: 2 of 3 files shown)


hector-cao:
@BFuhry Please take a look and give feedback

hector-cao:
> Overall, I consider it a good idea to prevent name collisions. But if I understand this change correctly, this will lead to the file names `/etc/apt/preferences.d/kobuk-tdx-kobuk-team-tdx-release-pin-4000` and `/etc/apt/preferences.d/kobuk-tdx-kobuk-team-tdx-attestation-release-pin-4000`. Why would you want/need `kobuk` in the name twice?
> 
> Related topic: would it make sense to have different names for the actual packages inside the package repository?

The first kobuk is to specify that this configuration file has been generated by the kobuk setup scripts.
The second kobuk is part of the Launchpad PPA team name, it can be kobuk-xxx or something else if the PPA does not belong to kobuk team

BFuhry:
> > Overall, I consider it a good idea to prevent name collisions. But if I understand this change correctly, this will lead to the file names `/etc/apt/preferences.d/kobuk-tdx-kobuk-team-tdx-release-pin-4000` and `/etc/apt/preferences.d/kobuk-tdx-kobuk-team-tdx-attestation-release-pin-4000`. Why would you want/need `kobuk` in the name twice?
> > Related topic: would it make sense to have different names for the actual packages inside the package repository?
> 
> The first kobuk is to specify that this configuration file has been generated by the kobuk setup scripts. The second kobuk is part of the Launchpad PPA team name (https://launchpad.net/~kobuk-team), it can be kobuk-xxx or something else if the PPA does not belong to kobuk team

Sure, makes sense

setup-tdx-guest.sh
```diff
@@ -56,7 +56,10 @@ apt update
 apt install --yes software-properties-common gawk &> /dev/null
 
 # cleanup
-rm -f /etc/apt/preferences.d/*tdx-*
+# NB: '*' before kobuk to keep backward compatiblity to make sure
+# we clean up all conf files that have been deployed in the
+# previous releases
+rm -f /etc/apt/preferences.d/*kobuk*tdx-*
```
ThibF, line 62: So we will never clean up existing config that doesn't start have kobuk before tdx, is that okay ?

hector-cao: It does, that is the reason why I proposed to add `*` before `kobuk`, to clean up conf files without kobuk as prefix but has only kobuk in their naming

ThibF (approved):

PR #403, opened 2025-08-22
Fix setup_ppcie to prepare gpus and switches.

Evrard-Nil:
- Added the -D flag to lspci to ensure the full domain:bus:device.function format is returned
- Fixed if condition: `len(args.gpus) == 8` to `len(gpus) == 8` as args.gpus is a string

### DIFF
--- a/guest-tools/run_td
+++ b/guest-tools/run_td
@@ -206,12 +206,24 @@ def run_td(args):
        return
     if args.image:
         td_img=args.image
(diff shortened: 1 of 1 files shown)


tmagik:
This appears to work better than my version which does not set up vfio correctly after the host is rebooted.

Evrard-Nil:
@tmagik can we get this merge?

PR #406, opened 2025-09-04
Fix the 8 gpu PPCIE mode detection logic

tmagik:
The code to automatically set PPCIE mode used the original string instead of the parsed and split gpu list, which results in the code never actually running when it's supposed to.


hector-cao (approved):
LGTM, thanks

